PR #1823, opened 2015-12-01
Add multithreaded, multipart upload support

earlephilhower:
Howdy Jason,

As promised, here is the updated code.  The diffs are mostly related to multipart/multithread handling in the upload stage, but to do that required removing the global status variable.  So you'll see lots of "unrelated" function edits.  It also needs the latest libs3 to be pulled in (or remove libs3 from irods directories completely if there's no real main trunk need).

A minor addition to hostname handling is the support of multiple hosts, comma separated.  This is needed for some S3 appliances where you have multiple control heads (but a consistent view of things behind the scenes).  Amazon has HW load balancers on s3.amazonaws.com to do this transparently to the user, but this allows for explicit SW load balancing for appliances w/o that HW in front of them.  If only one host (or none) are specified, there is no change in behavior.

The test .py file now allows environment overrides, but defaults to your original settings so it should not affect your CI system at all.  This allows me to run against my own test S3 appliance w/o patching the code.

Other stuff was just cleanup:

There were 2 structs{} defined as callback values in the header file, and both were used by callback functions....but only the first one was actually created/written to by the main function.  Because the first few variables matched in size and order it worked....until I had to start adding values to the struct and alignments shifted.  I cleaned that up and worked through things to ensure every callback is passed the proper struct type it expects, and no "lucky" aliasing is needed.

There was also some dead code.  A couple #ifdef (linux_platform ) (which actually never was defined in the build) could go, as well as many spots where the S3 operation functions tried to get the s3_hostname, but never actually used it.

Thanks,
-EFP3

### DIFF
--- a/README
+++ b/README
@@ -43,3 +43,5 @@ To control multipart uploads, add the resource variables "S3_MPU_CHUNK" and "S3_
 To ensure end-to-end data integrity, MD5 checksums can be calculated and used for S3 uploads.  Note that this requires 2x the disk IO (because the file must first be read to calculate the MD5 before the S3 upload can start) and a corresponding increase in CPU usage
 S3_ENABLE_MD5=[0|1]  (default is 0, off)
 
+S3 server side encryption can be enabled using the parameter S3_SERVER_ENCRYPT=[0|1] (default=0=off).  This is not the same as HTTPS, and implies that the data will be stored on disk encrypted.  To encrypt during the network transport to S3, please use S3_PROTO=HTTPS (the default)
+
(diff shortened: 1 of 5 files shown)


trel:
Thanks Earle, this looks really good.  We'll run it through our CI and confirm.

earlephilhower:
I"m going to close this one and send a new pull request.  Overnight one of our partners suggested that MD5 checksums and encryption would be required for some users, so I've added those in and run through the suite to check 'em out.  Small deltas to the original pull request.

PR #1828, opened 2016-03-02
Add support to enable use of Date header in S3 StringToSign

beppodb:
This is for compatibility with some non-Amazon S3-based services that use the Date header in the StringToSign, which is signed with the user's key and used for authentication.

See http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html "Time Stamp Requirement" and "Example Delete"

This request depends upon a concurrent libs3 pull request.

Usage:
```
... context="...;S3_STSDATE=DATE"
... context="...;S3_STSDATE=BOTH"
``` 

Using "DATE" includes the timestamp only in the Date header.
Using "BOTH" includes the timestamp in both the Date header and the x-amz-date header.
Any other value results in the default behavior, which includes the timestamp only in the x-amz-date header.


trel:
depends on https://github.com/irods/libs3/pull/1

adetorcy:
Manually merged: 9ddc7f655a5f258d0a2b2242cc4c47c0bd6f812e

PR #1832, opened 2016-06-03
Multipart, parallel copy of objects

earlephilhower:
The prior multipart/multithread patch did not use a multipart copy operation, and would fail on Amazon objects of > 5GB in size.  This patch adds that capability automatically for objects which are larger than the multipart chunk size.

It's much smaller a change than it looks because there are a lot of whitespace diffs because of the "if(OK) { if (OK) { if (OK) { ... } } } } } return status;" coding style it started with.    The whole s3CopyObject() changes are only maybe 5 lines, the rest are indents due to an add'l if() level needed for the object size read.

You will need the latest libs3 from yesterday with the new S3_copy_object_range() API added to compile and run successfully.

Also demoted some minor performance stats to DEBUG log level to keep chattiness down.


adetorcy:
Thanks!

PR #1840, opened 2018-04-11
Changes and documentation to use this plugin with Google Cloud Storage

HowardLander:


HowardLander:
One thing I didn't do is apply a fix from the gcs cloud storage test to the s3 test file. I don't have any s3 account credentials so I had no way to test that.  Here's that fix:
Change the line
import boto3

to

try:
   import boto3
except ImportError:
   print('This test requires boto3: perhaps try pip install boto3')
   exit()

trel:
Can you rebase and push to https://github.com/heliumdatacommons/irods_resource_plugin_s3 again?

PR #1859, opened 2019-04-13
Issue 1858 4 2 stable

JustinKyleJames:


s3/libirods_s3.cpp
```diff
@@ -226,13 +228,23 @@ static char *s3CalcMD5( int fd, off_t start, off_t length )
...
+
+    std::vector<std::string> hostname_vector;
+    size_t hostname_index = 0;
+    _prop_map.get<std::vector<std::string> >(s3_default_hostname_vector, hostname_vector);
+    _prop_map.get<size_t>(s3_hostname_index, hostname_index);
+    if (hostname_vector.empty()) {
         return NULL; // Short-circuit default case
+    }
+
     g_hostnameIdxLock.lock();
```
alanking: Is the lock needed? The hostname index is retrieved/set elsewhere without a lock, so either they need the lock as well or it is not needed at all.

s3/libirods_s3.cpp
```diff
@@ -494,121 +506,126 @@ irods::error s3ReadAuthInfo(
...
         }
+    }
+    _prop_map.set<size_t>(s3_retry_count_size_t, retry_count);
 
-        size_t ctr = 0;
-        while( ctr < g_retry_count ) {
-            int status = 0;
-            int flags = S3_INIT_ALL;
-            S3SignatureVersion signature_version = s3GetSignatureVersion(_prop_map);
+    size_t wait_time = 1;
```
alanking: Should we use `S3_DEFAULT_RETRY_WAIT_SEC` here?

s3/libirods_s3.cpp
```diff
@@ -494,121 +506,126 @@ irods::error s3ReadAuthInfo(
...
+    std::string wait_time_str;
+    ret = _prop_map.get< std::string >(
+        s3_wait_time_sec,
+        wait_time_str );
+    if( ret.ok() ) {
+        try {
+            wait_time = boost::lexical_cast<int>( wait_time_str );
```
alanking: `wait_time` is a `size_t` but we are lexical casting to `int` (and then implicitly back to `size_t`). What happens when the value is negative? Should we use a `decltype(wait_time)` or simply a `size_t` in the template parameter for consistency?

s3/libirods_s3.cpp
```diff
@@ -494,121 +506,126 @@ irods::error s3ReadAuthInfo(
...
-                LOG_NOTICE,
-                "s3Init - Error in connection, retry count %d",
-                ctr );
+        ctr++;
 
-        } // while
+        s3_sleep( wait_time, 0 );
 
-        if( result.ok() ) {
-            S3Initialized = true;
```
alanking: Can this global be removed now? It is initialized to false and only checked in one other place which creates a dead codepath.

s3/libirods_s3.cpp
JustinKyleJames: Good point.  I don't believe I changed/removed the lock in the other places.  My guess is they didn't have it in the init function because that was guaranteed to be executed by only one thread at a time.  (?).

Also that lock should protect the hostname_vector as well.  We read the size of the vector outside of any lock.  Will change.

s3/libirods_s3.cpp
JustinKyleJames: That would be a change to the interface as this is specified as (S3_WAIT_TIME_SEC) is in the context string.

Frankly, I don't think S3_DEFAULT_HOSTNAME should have DEFAULT in the name.  It isn't the default.  It is what we are setting it to.  But, again, changing it would be a change to the interface.

s3/libirods_s3.cpp
JustinKyleJames: Not changed code (other than indention) but will fix.

s3/libirods_s3.cpp
JustinKyleJames: Added back in due to its need with S3_deinitialize().

s3/libirods_s3.cpp
```diff
@@ -226,13 +228,22 @@ static char *s3CalcMD5( int fd, off_t start, off_t length )
...
 {
-    if (g_hostname.empty())
+
+
+    std::vector<std::string> hostname_vector;
+    size_t hostname_index = 0;
+    _prop_map.get<std::vector<std::string> >(s3_default_hostname_vector, hostname_vector);
+    _prop_map.get<size_t>(s3_hostname_index, hostname_index);
+    if (hostname_vector.empty())
         return NULL; // Short-circuit default case
```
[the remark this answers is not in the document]
JustinKyleJames: done

packaging/test_irods_resource_plugin_s3.py
```diff
@@ -459,6 +459,76 @@ def test_decoupled_naming_policy_issue1855(self):
...
+        # make sure we have four replicas
+        session.assert_icommand('ils -L %s' % filename, 'STDOUT_MULTILINE', ['s3repl_1858;s3compound1_1858;s3cache1_1858',
+                                                                             's3repl_1858;s3compound1_1858;s3archive1_1858',
+                                                                             's3repl_1858;s3compound2_1858;s3cache2_1858',
+                                                                             's3repl_1858;s3compound2_1858;s3archive2_1858'])
+
+        # remove the file
+        session.assert_icommand('irm -f %s' % filename) # remove file
+
+        # cleanup
```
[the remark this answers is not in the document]
JustinKyleJames: done

JustinKyleJames:
This has been superseded by pull request #1876.

Issue #1886, opened 2019-08-12
Update README to instruct user to use 4-2-stable branch.


JustinKyleJames:
Update README to instruct user to use 4-2-stable branch.

PR #1888, opened 2019-08-12
[#1886] Update README with instructions to use 4-2-stable

JustinKyleJames:


trel:
consider the `git commit -b 4-2-stable` notation as more compact?

JustinKyleJames:
I made changes to simplify the build instructions by using "git clone -b".

PR #1902, opened 2019-10-02
Issue 1872 master

JustinKyleJames:


JustinKyleJames:
Note the one Codacy error is not valid.  It is being used and if that line is deleted it won't compile.

PR #1904, opened 2019-10-16
[1903] Fixed the stomping of error message (4-2-stable)

JustinKyleJames:
Fixed the stomping of the detailed error message when the keypair file is not found.


trel:
if this is passing - yes, please #

PR #1911, opened 2020-01-04
Expanded some areas of the README in an effort to add additional clarity

yeea:
Happy New Year! 

Hopefully these changes are sensible to you guys. I tried to expand a bit on the expected configuration of the resource as well as points that may be a little confusing to other new users.


trel:
Seems good - can you squash them to a single commit?

No worries if not - but if I squash, then I believe your name won't stay attached as the committer.

yeea:
Thanks for the heads up; I've squashed the commits!

PR #1916, opened 2020-03-19
[#1806] Remove unused static functions from s3fs (4-2-stable)

alanking:
-Wunused-function trips on several static functions in s3fs which
have now been removed.

---

Need to run tests.


alanking:
CI tests passed

PR #1923, opened 2020-04-28
[irods/irods4860] Updated test cases for warning message

JustinKyleJames:
about specifying minimum number of replicas is deprecated


trel:
if these are passing, please # and then get onto master as well.

PR #1925, opened 2020-05-01
Getting master up to 4-2-stable

JustinKyleJames:
Getting master up to 4-2-stable.

Differences between master and 4-2-stable:

* Test code change for irods/irods#4860 is not necessary in 4.3.
* Change to S3 plugin operation signatures (additions of const).
* Change in version in CMakeLists.txt.

This passed locally run tests.


trel:
i like it - let's make the last commit point to 1806 as well.

Issue #1963, opened 2021-05-24
minio-py 7.0.x not compatible with current test suite


alanking:
- [x] master
- [x] 4-2-stable
---
The latest major version of `minio-py` is not compatible with the test suites. Identified issues include:
 - Removal of `ResponseError` object from `minio.errors` module
 - Use of `urllib.parse` (introduced in python3, which is not currently supported in the test-running script)

There may be more, but this is what I saw when trying to run the test suite through CI

JustinKyleJames:
This has been merged to master.  It can be closed.

PR #1964, opened 2021-05-24
[#1963] Use minio-py version 6.0.2 (4-2-stable)

alanking:
minio-py 7.0.x is not compatible with python2, which is the version of
python used for running the test suites at this time. Lock into the
latest version of the last major version number, which is 6.0.2 until
this is resolved.

Some CI tests are failing, but importing the `minio` module is not the cause. I believe the 3 tests that I see failing are due to some changes in the behavior of the replication API in the server


trel:
looks good to me - # when satisfied/passing

alanking:
Failing tests are now passing. CI test suite is still running, but I went ahead and added #

PR #1967, opened 2021-06-01
4 2 stable working

JustinKyleJames:
Pull with the latest work.  This is passing test in CI.


README.md
```diff
@@ -65,24 +65,23 @@ If the `S3_DEFAULT_HOSTNAME` points to an AWS host, [best practice](https://docs
...
 ```
 
-To control multipart uploads:
--   `S3_MPU_CHUNK` is the size of each part to be uploaded in parallel (in MB, default is 5MB).  Objects smaller than this will be uploaded with standard PUTs.
--   `S3_MPU_THREADS` is the number of parts to upload in parallel (only under Linux, default is 10).  On non-Linux OSes, this parameter is ignored and multipart uploads are performed sequentially.
-*The above settings do not apply to multipart uploads in streaming mode.  In this case the behavior is determined by the parallel transfer configuration in iRODS.*
+To define S3 provider constraints and control multipart behavior:
+-    `S3_MPU_CHUNK` - This defines the minimum part size allowed (in MB).  The default is 5MB which is the minimum part size defined in AWS.
+-    `S3_ENABLE_MPU=0` disables multipart uploads.
+-    `S3_MAX_UPLOAD_SIZE` - This defines the maximum upload size for non-multipart uploads (in MB), maximum part size, as well as the maximum size when using the CopyObject API.  The default is 5120MB (5GB).  This setting is ignored if MPU uploads are disable.
```
trel, line 71: 'disable' -> 'disabled'

s3/libirods_s3.cpp
```diff
@@ -906,6 +895,90 @@ S3STSDate s3GetSTSDate( irods::plugin_property_map& _prop_map)
...
+
+        // if cachedir is defined, use that else use /tmp/<resc_name>
+        std::string s3_cache_dir_str;
+        irods::error ret = _prop_map.get< std::string >(s3_cache_dir, s3_cache_dir_str);
+        if (!ret.ok()) {
+            s3_cache_dir_str = boost::filesystem::temp_directory_path().string();
+        }
+
+        // No longer using shared_memory_name_salt because it uses irods main server process PID and a hash
+        // which changes every time the server is restarted.  While that is preferrable in case something goes
```
trel, line 908: 'preferrable' -> 'preferable'

s3/s3_operations.cpp
```diff
@@ -120,20 +123,103 @@ namespace irods_s3 {
             int fd_counter;
     };
 
-    int debug_log_level = LOG_DEBUG;
+    int debug_log_level = LOG_NOTICE;
```
trel, line 126: is this going to be too verbose for production deployments?

s3/s3_transport/include/s3_transport.hpp
```diff
@@ -932,11 +1044,13 @@ namespace irods::experimental::io::s3_transport
 
                 // override for cases where we must use cache
                 //   1. If we don't know the file size.
-                //   2. If doing multiplart upload file size < #threads * minimum part size
+                //   2. If we have > 1 thread and multipart is disabled
+                //   3. If doing multiplart upload file size < #threads * minimum part size
```
trel, line 1048: 'multiplart' -> 'multipart'

s3/s3_transport/unit_tests/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(unit_tests LANGUAGES C CXX)
 
 find_package(IRODS 4.2.8 EXACT REQUIRED)
```
trel, line 4: 4.2.9 also?

trel:
typos in the commit message for 107ab1422441ea656109f0f8dbff4bcf88708b0d

update to this:
```
Now that we are pushing parts that may be close to the circular buffer
size, there was a potential deadlock situation that could arise:

1.  The thread pushing on the circular buffer is trying to push a buffer
that exceeds the buffer size.
2.  It queries the circular buffer to see how many it can push.
3.  Between this query more items are pushed and therefore it tries to
push beyond the circular buffer size.

To fix this, the query for how many can be pushed needs to be in the
circular buffer itself so that it is inside the lock.  The push_back()
returns the number of bytes pushed so the caller can determine via this
information if it needs to push more.
```

README.md
JustinKyleJames: done

s3/libirods_s3.cpp
JustinKyleJames: done

s3/s3_operations.cpp
JustinKyleJames: This is correct.  I missed this setting.  Updated.

s3/s3_transport/include/s3_transport.hpp
JustinKyleJames: fixed

s3/s3_transport/unit_tests/CMakeLists.txt
JustinKyleJames: fixed

s3/s3_operations.cpp
```diff
@@ -377,7 +377,7 @@ namespace irods_s3 {
         }
 
         // ********* DEBUG - print L1desc for all
-        if (debug_log_level == LOG_DEBUG) {
+        if (debug_log_level < LOG_DEBUG) {
```
trel, line 380: you want this to print if debug_log_level is DEBUG or noisier, right?

so...  i think that means this should be `>= LOG_DEBUG`?

JustinKyleJames: Normally the debug_log_level will be LOG_DEBUG (7).  We do not want this loop to run at this level.  To turn on more verbose messages the debug_log_level would be set to LOG_NOTICE (5).  At that point we would want to run this loop for debugging purposes.

trel: phone call - decided on only printing this loop out if server properties log level is >= this local developer log level.

PR #1986, opened 2021-06-23
[1880] readme changes for path request style

JustinKyleJames:


trel:
this seems good to me - and yes on the backwards compatibility.  please # if you're happy with it.

PR #1994, opened 2021-08-05
[1993] update google cloud storage section of README

trel:


JustinKyleJames:
After testing I updated the README as follows.  Maybe merge this with this pull request?

The notes about https, url, and bucket seem to be just normal configuration not exception case.  We could keep it in there but maybe under a different section.  ?

## Using this plugin with Google Cloud Storage (GCS)

This plugin has been manually tested to work with Google Cloud Storage with some caveats.
-   If multipart uploads are enabled (S3_ENABLE_MPU=1), the call to CopyObject gives an error.  Therefore, if MPU is enabled then CopyObject should be disabled via  S3_DISABLE_COPYOBJECT=1.
-   The values in the `S3_AUTH_FILE` have to be generated according to the instructions: https://cloud.google.com/storage/docs/migrating#keys
-   GCS treats bucket names with dots as domain names.  These must be verified.  See [Domain-named Bucket Verification] (https://cloud.google.com/storage/docs/domain-name-verification).

PR #2027, opened 2022-02-14
[irods/irods#3727] Remove ASSERT_* (and bzero) usage (main)

alanking:
Tests need to be run

### DIFF
--- a/irods_consortium_continuous_integration_test_hook.py
+++ b/irods_consortium_continuous_integration_test_hook.py
@@ -1,98 +1,119 @@
 from __future__ import print_function
 
+import glob
 import optparse
 import os
-import shutil
-import glob
-import time
 import random
+import shutil
+import stat
 import string
 import subprocess
+import time
 
 import irods_python_ci_utilities
 
-
-def get_build_prerequisites_apt():
-    return[]
-
-def get_build_prerequisites_yum():
-    return[]
-
-def get_build_prerequisites_zypper():
-    return[]
-
-
-def get_build_prerequisites():
-    dispatch_map = {
-        'Ubuntu': get_build_prerequisites_apt,
-        'Centos': get_build_prerequisites_yum,
-        'Centos linux': get_build_prerequisites_yum
-    }
-    try:
-        return dispatch_map[irods_python_ci_utilities.get_distribution()]()
-    except KeyError:
-        irods_python_ci_utilities.raise_not_implemented_for_distribution()
-
-
-def install_build_prerequisites():
-    irods_python_ci_utilities.subprocess_get_output(['sudo', 'pip', 'install', 'boto3', '--upgrade'], check_rc=True)
-    irods_python_ci_utilities.subprocess_get_output(['sudo', 'pip', 'install', 'minio==6.0.2', '--upgrade'], check_rc=True)
-    irods_python_ci_utilities.subprocess_get_output(['sudo', '-EH', 'pip', 'install', 'unittest-xml-reporting==1.14.0'])
-    if irods_python_ci_utilities.get_distribution() == 'Ubuntu': # cmake from externals requires newer libstdc++ on ub12
-        if irods_python_ci_utilities.get_distribution_version_major() == '12':
-            irods_python_ci_utilities.install_os_packages(['python-software-properties'])
-            irods_python_ci_utilities.subprocess_get_output(['sudo', 'add-apt-repository', '-y', 'ppa:ubuntu-toolchain-r/test'], check_rc=True)
-            irods_python_ci_utilities.install_os_packages(['libstdc++6'])
-
-    #irods_python_ci_utilities.install_os_packages(get_build_prerequisites())
+def install_test_prerequisites():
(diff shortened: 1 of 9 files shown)


alanking:
companion PR to irods/irods#6155

alanking:
Putting this back in draft. I do'nt think I actually finished it

trel:
#2041 is now merged

alanking:
Confirmed that this builds, haven't run tests yet.

alanking:
The test hook and suite need some work in order to run with python3 for main, so I will open a separate PR based on this branch for that. Unfortunately, main no longer builds because of use of the deprecated ASSERT_* macros, so we need to get this in first. I'll update once I get tests to run

alanking:
Okay, went with a different plan :) The latest force-push should be treated as a brand new set of changes, because they are.

In addition to removing ASSERT_* from the codebase, this changeset allows the test suite to be run in parallel in the testing environment. I have shown that 3 of the 6 test suites in `test_irods_resource_plugin_s3_minio` pass when run in parallel (for some reason) but all of them pass when run serially. I will create a new issue for this. I'm running the test suite one more time to make sure.

irods_consortium_continuous_integration_test_hook.py
```diff
@@ -81,6 +55,7 @@ def main():
     os_specific_directory = irods_python_ci_utilities.append_os_specific_directory(built_packages_root_directory)
 
     irods_python_ci_utilities.install_os_packages_from_files(glob.glob(os.path.join(os_specific_directory, 'irods-resource-plugin-s3*.{0}'.format(package_suffix))))
+    install_test_prerequisites()
 
     install_build_prerequisites()
```
alanking: I'm considering squashing the changes having to do with the test hook into one because separating them is hard. As you can see, this commit is not a valid change by itself as this function is deleted above.

alanking:
I confirmed that all of the tests passed. There was an interaction between the tests which caused some to fail, but I've captured this in #2057

JustinKyleJames (approved):
This all looks straight-forward.

s3/libirods_s3.cpp
```diff
@@ -431,106 +433,125 @@ irods::error parseS3Path (
...
     std::string resource_name = get_resource_name(_prop_map);
-
+    // TODO: why do we not return immediately? key_id and acccess_key are not referenced
+    // again, and there is no failure in the event that either key is missing from the
+    // environment.
     if ((tmpPtr = getenv(s3_key_id.c_str())) != NULL) {
         key_id = tmpPtr;
         if ((tmpPtr = getenv(s3_access_key.c_str())) != NULL) {
             access_key = tmpPtr;
         }
```
JustinKyleJames, line 515: Agreed.  I am not sure what the point of this is.  Why are environment variables being used?  And why set key_id and access_key if they are not going to be used?

If I understand this correctly, if the environment variables are set, they take precedence which would mean that you could not have multiple key pairs for multiple resources.

I would say torch this part.

s3/libirods_s3.cpp
```diff
@@ -610,16 +630,20 @@ irods::error s3InitPerOperation (
...
             }
 
+            // TODO: This will never be true as status is not set in this compound conditional
+            // statement whose predicate is status == S3StatusOK.
             if (status != S3StatusOK) {
                 std::string error_str =  boost::str(boost::format("[resource_name=%s] failed to set region name to %s: %s") % resource_name.c_str() %
                     region_name.c_str() % S3_get_status_name(status));
                 rodsLog(LOG_ERROR, error_str.c_str());
                 return ERROR(S3_INIT_ERROR, error_str.c_str());
             }
```
JustinKyleJames, line 641: Agreed on status.  In addition, we are not setting the region name in this routine so the "failed to set region name" error message wouldn't apply anyway.  I think we should torch this part.

The check above is simply to log a warning that we are using the default region.  After that we should just return SUCCESS().

JustinKyleJames (changes requested):
I had a couple of comments where I agreed with the comments in the code.  Please remove those code sections as requested.

s3/libirods_s3.cpp
alanking: I guess the question is whether or not we plan to support environment variables as an alternative to the keypair file. As it stands, the environment variable use case is not being fully supported, but it looks like it may have been in the past.

I'm okay with torching it but just know that we will be officially dropping that use case instead of having a broken feature (in which case I will file an issue). Thoughts?

s3/libirods_s3.cpp
alanking: Should I change the log message on 630 from LOG_ERROR to LOG_WARNING?

I will remove 633 through 641.

s3/libirods_s3.cpp
alanking: After discussing, we have decided that we want to drop this. I will do so in the next commit

s3/libirods_s3.cpp
alanking: Will do

alanking:
Addressed the comments!

irods_consortium_continuous_integration_test_hook.py
alanking: I will squash down the test hook-related changes

alanking:
Squashed and #'d and mergin

PR #2030, opened 2022-03-03
Issue 2024 (4-2-stable)

JustinKyleJames:
1. Fixed int64_t conversion to unsigned int in s3_upload_part_worker_routine which was causing send threads to bail
early when thread handles > 4 GiB of data.

2. Changed flush cache file code to make sure that each part <= 4 GiB.  There was previously a one-to-one relationship on threads and parts.  That remains the case as long as each part <= 4 GiB.  Otherwise the file is split into enough parts so that each are <= 4 GiB.  The threads handle the parts in a round-robin manner.

3. Fix shmem size for large number of parts.

4. Fixed malformed error message causing segfault when circular buffer writer times out.

5. Changed complete multipart and object delete to use an explicit timeout value.  With libs3 changes this will limit timeouts to this setting and the low data rate timeout will not apply.  This is required because these may take a long time and it sends very little data.

6. Added some additional debug logging.


trel:
this feels like it should be... 6-7 different commits?     perhaps against 5-6 different issues?

s3/s3_operations.cpp
```diff
@@ -1056,7 +1058,7 @@ namespace irods_s3 {
         S3_delete_object(
             &bucketContext,
             key.c_str(), 0,
-            0,                    //timeout
+            1800000,                    //timeout
```
alanking, line 1061: Should we capture this in a constant somewhere? Maybe even a config?

s3/s3_transport/include/managed_shared_memory_object.hpp
```diff
@@ -6,6 +6,8 @@
 #include <boost/interprocess/managed_shared_memory.hpp>
 #include <boost/container/scoped_allocator.hpp>
 
+#include <rodsLog.h>
```
alanking, line 9: I think this should be in quotation marks instead of brackets

s3/s3_transport/include/s3_transport.hpp
```diff
@@ -105,6 +105,7 @@ namespace irods::experimental::io::s3_transport
             , resource_name{""}
             , restoration_days{S3_DEFAULT_RESTORATION_DAYS}
             , restoration_tier{S3_DEFAULT_RESTORATION_TIER}
+            , max_single_part_upload_size{5L*1024*1024*1024}
```
alanking, line 108: Should we capture this in a constant?

s3/s3_transport/include/s3_transport.hpp
```diff
@@ -1003,23 +1038,52 @@ namespace irods::experimental::io::s3_transport
                 ? config_.number_of_cache_transfer_threads
                 : cache_file_size / minimum_part_size == 0 ? 1 : cache_file_size / minimum_part_size;
 
-            int64_t part_size = cache_file_size / config_.number_of_cache_transfer_threads;
+            // Calculate the number or parts.  This is usually a one-to-one mapping to number of threads but if
```
alanking, line 1041: "number or parts" -> "number of parts"

s3/s3_transport/include/s3_transport.hpp
```diff
@@ -1868,23 +1953,25 @@ namespace irods::experimental::io::s3_transport
...
+                        } catch (boost::interprocess::bad_alloc &biba) {
                             this->set_error(ERROR(S3_PUT_ERROR, "Error on reallocation of etags buffer in shared memory."));
                             data.last_error_code = error_codes::BAD_ALLOC;
                             return true;
                         }
+
                     }
 
+
                     return false;
```
alanking, line 1976: Would it make sense to move this lambda into a method since it's used in multiple places?

alanking (commented):
Agreed, breaking up the second commit with 1 commit and/or issue for each of the 6 bullet points (or maybe just the first 5?) would be good.

Also, I haven't been following this plugin very closely up to this point so my comments may be coming from a place of ignorace :)

packaging/test_irods_resource_plugin_s3_minio.py
```diff
@@ -92,3 +125,13 @@ def __init__(self, *args, **kwargs):
...
         self.s3EnableMPU=1
         super(Test_S3_NoCache_EU_Central_1, self).__init__(*args, **kwargs)
+
+    # issue 2024
+    @unittest.skipIf(True, "File removal too slow with MinIO")
+    def test_put_get_file_greater_than_8GiB_two_threads(self):
+        Test_S3_NoCache_Base.test_put_get_file_greater_than_8GiB_two_threads(self) 
+
+    # issue 2024
+    @unittest.skipIf(psutil.disk_usage('/').free < 4 * (4*1024*1024*1024 + 2), "not enough free space for three 4 GiB files (upload, download, and two on-disk MinIO)")
```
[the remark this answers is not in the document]
JustinKyleJames: yes, will change

s3/s3_transport/include/managed_shared_memory_object.hpp
JustinKyleJames: Why?  Quotes search for local directory first.  Brackets does not.  Seems like brackets is better?

JustinKyleJames: We discussed this.  It will be left as <rodsLog.h> for now.  Once Markus reorganizes the repositories this will be <irods/rodsLog.h>.

s3/s3_transport/include/s3_transport.hpp
JustinKyleJames: There should have only been one as I moved it up to run on both cache and cacheless flows.  Removing the second one as it is unnecessary.

packaging/test_irods_resource_plugin_s3_minio.py
```diff
@@ -92,3 +125,13 @@ def __init__(self, *args, **kwargs):
         self.s3endPoint='localhost:9001'
         self.s3EnableMPU=1
         super(Test_S3_NoCache_EU_Central_1, self).__init__(*args, **kwargs)
+
+    # issue 2024
+    @unittest.skipIf(True, "File removal too slow with MinIO")
```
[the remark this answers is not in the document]
JustinKyleJames: Done

packaging/test_irods_resource_plugin_s3_minio.py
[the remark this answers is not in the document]
JustinKyleJames: done

s3/s3_operations.cpp
JustinKyleJames: It is now a configuration parameter with a default that is a constant.

s3/s3_transport/include/s3_transport.hpp
JustinKyleJames: Done

s3/s3_transport/include/s3_transport.hpp
JustinKyleJames: Done

JustinKyleJames:
> this feels like it should be... 6-7 different commits? perhaps against 5-6 different issues?

I now have multiple issues and commits.

packaging/test_irods_resource_plugin_s3_minio.py
```diff
@@ -43,6 +46,16 @@ def __init__(self, *args, **kwargs):
         self.s3EnableMPU=1
         super(Test_S3_NoCache_V4, self).__init__(*args, **kwargs)
 
+    # issue 2024
+    @unittest.skipIf("File removal too slow with MinIO")
```
alanking, line 50: `skipIf` -> `skip`

Same for 70, 90, 94, 98, 102, 107, and 130.

README.md
```diff
@@ -89,7 +89,8 @@ To encrypt during the network transport to S3, use `S3_PROTO=HTTPS` (the default
 
 The `S3_RETRY_COUNT` defines the number of retries for a request after a retryable failure.  The default is 3 retries.
 The `S3_WAIT_TIME_SEC` defines the initial number of seconds between retries.  The default is 2s.  This wait time will double on each successive failure until the `S3_MAX_WAIT_TIME_SEC` is reached.
-The `S3_MAX_WAIT_TIME_SEC` is the maximum wait value during successive failures.  The default is 30s. 
+The `S3_MAX_WAIT_TIME_SEC` is the maximum wait value during successive failures.  The default is 30s.
+The `S3_NON_DATA_TRANSFER_TIMEOUT_SECONDS` defines the timeout value used for S3_complete_multipart_upload and S3_delete_object.  The default is 300s.  (Note:  This has been added because in some cases with very large files these take a long time to complete and the data transfer thresholds do not apply to these.)
```
alanking, line 93: Just curious: What happens if this value is 0? It looks like this is what we had hard-coded before, so I'm assuming it indicates no timeout. Would that be worth mentioning here?

s3/s3_transport/include/s3_transport.hpp
```diff
@@ -582,11 +582,27 @@ namespace irods::experimental::io::s3_transport
...
                 if ( use_streaming_multipart() ) {
-                    begin_part_upload_thread_ptr_ = std::make_unique<std::thread>(
-                            &s3_transport::s3_upload_part_worker_routine, this, false, 0, 0, get_file_offset());
+                    try {
+                        begin_part_upload_thread_ptr_ = std::make_unique<std::thread>(
+                                &s3_transport::s3_upload_part_worker_routine, this, false, 0, 0, get_file_offset());
+                    } catch (const std::bad_alloc& ba) {
+                        this->set_error(ERROR(S3_PUT_ERROR, "Allocation error when creating upload part thread."));
+                        return 0;
+                    } catch (const std::system_error& se) {
```
alanking, line 591: `ba` and `se` are unused. Should they be part of the error message (like on 805 below)?

alanking (commented):
The different commits helped, thanks for doing that

README.md
JustinKyleJames: Mistake on my part.  Will fix.

JustinKyleJames: If the value is 0 it will simply not wait and retries will begin immediately.  This and S3_WAIT_TIME_SEC are different from the others as it isn't a transport type timeout.  These are part of the exponential backoff.

I could state this but I think it is implied.

trel: please write any/all of that down.   leave nothing implied - this s3 plugin already has many complex interactions, so making them explicit will be good for everyone.

s3/s3_transport/include/s3_transport.hpp
```diff
@@ -1980,6 +1980,8 @@ namespace irods::experimental::io::s3_transport
 
             for (unsigned int part_number = start_part_number; part_number <= end_part_number; ++part_number) {
 
+		retry_cnt = 0;
```
alanking, line 1983: Indentationnnn

README.md
alanking: I second this motion

s3/s3_transport/include/s3_transport.hpp
alanking: Looks done to me

s3/s3_transport/include/s3_transport_callbacks.hpp
```diff
@@ -619,30 +617,25 @@ namespace irods::experimental::io::s3_transport
...
+                            } else {
+                                data.etags[callback_for_write_to_s3_base_data->sequence - 1] = "";
                             }
-                        }
 
-                        if (etag) {
-                            data.etags[callback_for_write_to_s3_base_data->sequence - 1] = etag;
-                        } else {
-                            data.etags[callback_for_write_to_s3_base_data->sequence - 1] = "";
+                        } catch (bi::bad_alloc& ba) {
```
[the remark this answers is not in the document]
alanking: Looks like this is done in a future commit.

alanking (commented):
Almost there...

s3/s3_transport/include/s3_transport.hpp
alanking: Done in future commit

packaging/test_irods_resource_plugin_s3_minio.py
alanking: Looks done to me

s3/s3_transport/include/s3_transport.hpp
```diff
@@ -1988,7 +1988,7 @@ namespace irods::experimental::io::s3_transport
 
             for (unsigned int part_number = start_part_number; part_number <= end_part_number; ++part_number) {
 
-		retry_cnt = 0;
+                retry_cnt = 0;
```
trel, line 1991: squash with 2036

README.md
JustinKyleJames: This is done.

alanking: Does a value of 0 for the `S3_NON_DATA_TRANSFER_TIMEOUT_SECONDS` setting have the same meaning as a value of 0 for `S3_MAX_WAIT_TIME_SEC`?

trel:
got it - last two commits look good - please squash them where appropriate.

JustinKyleJames:
> > [2027] Updated code to make sure max upload size is never exceeded.
> 
> 2027 should be updated to point to the correct issue

This should be 2029.  It has been updated.

trel:
agreed - please # em!

JustinKyleJames:
> agreed - please # em!

Done.

trel:
I see no #'s yet.

PR #2062, opened 2022-04-05
Issue 2058

JustinKyleJames:


s3/s3_operations.cpp
```diff
@@ -168,22 +169,22 @@ namespace irods_s3 {
         std::uint64_t thread_id = std::hash<std::thread::id>{}(std::this_thread::get_id());
 
         // ********* DEBUG - print L1desc for all
-        if (getRodsLogLevel() >= developer_messages_log_level) {
-            rodsLog(developer_messages_log_level, "%s:%d (%s) [[%lu]] ------------- L1desc ---------------\n",
+        if (irods::experimental::log::logger_config<s3_plugin_logging_category>::level >= log::level::debug) {
```
korydraughn, line 172: Be careful checking the log level. It isn't thread-safe.

korydraughn (commented):
Just one comment.

Is there any risk of S3 log level being changed within the same agent?

alanking (commented):
I don't see s3_transport_logging_category.hpp or s3_plugin_logging_category.hpp anywhere.

s3/s3_operations.cpp
JustinKyleJames: The level only gets set on agent startup from the config in server_config.json.

The idea here was to skip the debug loop (over L1desc) if we are not logging debug logs.  Worst case as far as I can tell, if this test passed in error, we would iterate over L1desc but nothing would be logged because the actual log messages are set as debug.

s3/libirods_s3.cpp
```diff
@@ -582,14 +587,9 @@ irods::error s3InitPerOperation (
...
 
-    std::size_t retry_count = 10;
+    std::size_t retry_count = S3_DEFAULT_RETRY_COUNT;
     std::string retry_count_str;
     auto ret = _prop_map.get<std::size_t>(s3_retry_count, retry_count);
-    if (!ret.ok()) {
-        irods::log(LOG_NOTICE, fmt::format(
-                   "[resource_name={}] - Failed to get property [{}]",
-                   resource_name, s3_retry_count));
-    }
```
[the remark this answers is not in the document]
JustinKyleJames: Good catch.  I forgot to add them.  Will do.

s3/libirods_s3.cpp
```diff
@@ -943,24 +942,22 @@ std::size_t get_retry_wait_time_sec(irods::plugin_property_map& _prop_map) {
...
+                "[resource_name={}] failed to cast {} [{}] to a std::size_t", resource_name.c_str(),
                 s3_wait_time_seconds.c_str(), wait_time_str.c_str() );
         }
     } else {
         // for backward compatibility, look for S3_WAIT_TIME_SEC
         irods::error ret = _prop_map.get< std::string >( s3_wait_time_sec, wait_time_str );
         std::string resource_name = get_resource_name(_prop_map);
         if( ret.ok() ) {
-            irods::log(LOG_WARNING, fmt::format("[resource_name={} - {} is deprecated.  Use {}",
+            s3_logger::warn(fmt::format("[resource_name={} - {} is deprecated.  Use {}",
```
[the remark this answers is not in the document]
JustinKyleJames: Missed this one on search/replace.

s3/libirods_s3.cpp
```diff
@@ -1218,8 +1209,8 @@ irods::error s3GetFile(
 
         g_mrdData = (multirange_data_t*)calloc(totalSeq, sizeof(multirange_data_t));
         if (!g_mrdData) {
-            std::string msg =  boost::str(boost::format("[resource_name=%s] Out of memory error in S3 multirange g_mrdData allocation.") % resource_name.c_str());
-            rodsLog( LOG_ERROR, msg.c_str() );
+            auto msg =  fmt::format("[resource_name={}] Out of memory error in S3 multirange g_mrdData allocation.", resource_name);
+            s3_logger::debug("{} Out of memory error in S3 multirange g_mrdData allocation.",  msg);
```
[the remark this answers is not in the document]
JustinKyleJames: Fixed.

s3/s3_operations.cpp
korydraughn: Cool with me.

s3/s3_transport/include/s3_transport_logging_category.hpp
```diff
@@ -0,0 +1,34 @@
...
+    class log::logger_config<s3_transport_logging_category>
+    {
+        // This defines the name that will appear in the log under the "log_category" key.
+        // The "log_category" key defines where the message originated. Try to use a name
+        // that makes it easy for administrators to determine what produced the message.
+        static constexpr const char* name = "s3_transport_logging_category";
+
+        // This is the current log level for the category. This also represents the initial
+        // log level. Use the "set_level()" function to adjust the level.
+        static inline log::level level = log::level::info;
```
alanking, line 25: The log level in the plugin logger is explicitly declared as a public member. Is there a reason it is not here?

s3/s3_plugin_logging_category.hpp
```diff
@@ -0,0 +1,37 @@
...
+        // This is required since the fields above are private.
+        // This allows the logger to access and modify the configuration.
+        friend class logger<s3_plugin_logging_category>;
+
+    public:
+        // This is the current log level for the category. This also represents the initial
+        // log level. Use the "set_level()" function to adjust the level.
+        static inline log::level level = log::level::info;
+
+    };
```
alanking, line 33: @korydraughn, we have some convenience types declared in `irods_logger.hpp` which might be useful here. Do you think it would be appropriate to do so for these new categories somewhere in the header file?

https://github.com/irods/irods/blob/eb361f5e49e817ff498dbb417eb935030e5b2322/lib/core/include/irods/irods_logger.hpp#L74-L85

alanking (commented):
Looks good to me! I had a couple of questions but I don't think they're blockers by any means

s3/s3_plugin_logging_category.hpp
korydraughn: Are you saying to define them in `irods_logger.hpp` or this particular header file?

I think it is perfectly fine to define type aliases for the s3 log category in this file. It would make it a little more convenient to use the logger then.

The type alias should be defined outside of the experimental namespace unless it's truly experimental.

alanking: Right, I feel like this is more in line with what I see in irods/irods:
```c++
using s3_logger = irods::experimental::log::s3_plugin;
```
Which would act as a replacement for:
```c++
using s3_logger = irods::experimental::log::logger<s3_plugin_logging_category>;
```

alanking: We can argue about the naming, just wanted to throw out that the type alias would be nice to introduce here.

s3/s3_transport/include/s3_transport_logging_category.hpp
JustinKyleJames: I needed read access to it for the debug logging loop to print the L1desc table.  I didn't need this for the transport one.

Should I instead change it back to private with a getter message?

JustinKyleJames: Changed the plugin logger's level back to private and added a getter.

```
        // This is the current log level for the category. This also represents the initial
        // log level. Use the "set_level()" function to adjust the level.
        static inline log::level level = log::level::info;

    public:

        static auto get_level() -> decltype(level)
        {   
            return level;
        }  
```

korydraughn: We know the type of `level` here. Why not use `log::level` instead of `decltype(level)`?

s3/s3_plugin_logging_category.hpp
JustinKyleJames: Talked to Alan.  We determined that this might not be possible due to how this is set up.

JustinKyleJames:
Pushed code @alanking @korydraughn.  Please check again and let me know if I can squash/pound.

s3/s3_transport/include/s3_transport_logging_category.hpp
JustinKyleJames: Done

korydraughn (approved):
Looks good to me.

alanking (approved):
Let's squash em

JustinKyleJames:
> Let's squash em

Squashed.  No pound yet.

alanking (approved):
Looks good to me

korydraughn:
@JustinKyleJames Is there something else you want us to take a look at?

JustinKyleJames:
Pounds added.  @korydraughn @alanking

PR #2077, opened 2022-07-25
Issue 2025 - deep archive

JustinKyleJames:
This has several changes:

- Added support for deep archive
- Added support for setting the storage class in resource context
- Fixed parsing of restoration_days to detect invalid negative values
- Fixed itouch to S3 resource
- Fixed test case testing registration from S3 bucket

Note:  The libs3 version will need to be updated before we merge.


s3/s3_operations.cpp
```diff
@@ -1210,7 +1217,22 @@ namespace irods_s3 {
             return SUCCESS();
         }
 
+        // issue 1978 - If they called create without close, the object will not yet exist in s3.
+        // To be consistent with POSIX, return size of 0.
```
alanking, line 1221: Would we need to do any of this if `replica_close` closed the L3 descriptor before finalizing the replica?

In `rsDataObjClose`, we call `close_physical_file_and_throw_on_failure` before stat'ing the replica or doing anything with the catalog. In `replica_close`, we update the size and status of the replica in the catalog, and *then* close the physical file.

If the physical file was closed first, we wouldn't need to do this because the close operation would be called before we perform the stat and supposedly the object would exist and we could return the true size (which would be 0), right?

alanking (commented):
Glacier changes look good overall.

I have a concern about the `itouch` issue. I'm wondering if we need to make that fix in the server rather than in this plugin because I'm thinking that the plugin is not at fault here.

s3/s3_operations.cpp
alanking: For reference: https://github.com/irods/irods/blob/ed98d6d66e861f6ace4f72769e8f251856153f4f/plugins/api/src/replica_close.cpp#L431-L440

The suggestion is that perhaps we could move that block up above the block where the catalog updates occur: https://github.com/irods/irods/blob/ed98d6d66e861f6ace4f72769e8f251856153f4f/plugins/api/src/replica_close.cpp#L370-L371

Also for reference, here is the function in `rsDataObjClose`: https://github.com/irods/irods/blob/ed98d6d66e861f6ace4f72769e8f251856153f4f/server/api/src/rsDataObjClose.cpp#L586-L639 The very first thing that happens is that the physical file is closed.

alanking:
For my own reference, going to link the sibling PR: https://github.com/irods/libs3/pull/21

s3/libirods_s3.cpp
```diff
@@ -1098,26 +1108,59 @@ unsigned int s3_get_restoration_days(irods::plugin_property_map& _prop_map) {
...
+        return s3_transport::S3_STORAGE_CLASS_STANDARD;
+    } else if ( boost::iequals(storage_class_str.c_str(), s3_transport::S3_STORAGE_CLASS_GLACIER)) {
+        return s3_transport::S3_STORAGE_CLASS_GLACIER;
+    } else if ( boost::iequals(storage_class_str.c_str(), s3_transport::S3_STORAGE_CLASS_DEEP_ARCHIVE)) {
+        return s3_transport::S3_STORAGE_CLASS_DEEP_ARCHIVE;
+    } else if ( boost::iequals(storage_class_str.c_str(), s3_transport::S3_STORAGE_CLASS_GLACIER_IR)) {
+        return s3_transport::S3_STORAGE_CLASS_GLACIER_IR;
+    } else {
+        std::string resource_name = get_resource_name(_prop_map);
+        s3_logger::error(
```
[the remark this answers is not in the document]
JustinKyleJames: Warning is more appropriate.  Info is probably not enough as it wouldn't normally trigger someone to notice.  I am changing to warning.

s3/s3_operations.cpp
korydraughn: To build on top of this, we could provide an additional option on close that instructs it to swizzle the closing of the file on disk. Then the S3 resource can leverage that since it's special. Everything else can continue to operate as originally defined.

s3/s3_transport/unit_tests/CMakeLists.txt
```diff
@@ -21,11 +21,6 @@ endif()
 
 macro(IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH DEPENDENCY_NAME DEPENDENCY_SUBDIRECTORY)
   if (IS_DIRECTORY ${IRODS_EXTERNALS_FULLPATH_${DEPENDENCY_NAME}})
-    message(STATUS "Using user-specified value for IRODS_EXTERNALS_FULLPATH_${DEPENDENCY_NAME}: ${IRODS_EXTERNALS_FULLPATH_${DEPENDENCY_NAME}}")
-  else()
-    if (NOT IS_DIRECTORY ${IRODS_EXTERNALS_PACKAGE_ROOT}/${DEPENDENCY_SUBDIRECTORY})
-      message(FATAL_ERROR "${DEPENDENCY_NAME} not found at ${IRODS_EXTERNALS_PACKAGE_ROOT}/${DEPENDENCY_SUBDIRECTORY}")
-    endif()
```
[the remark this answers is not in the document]
JustinKyleJames: done

s3/libirods_s3.hpp
```diff
@@ -52,10 +52,19 @@ extern const std::size_t  S3_DEFAULT_MAX_RETRY_WAIT_SECONDS;
...
+namespace irods::experimental::io::s3_transport
+{
+    extern const unsigned int S3_DEFAULT_RESTORATION_DAYS;
+    extern const std::string  S3_DEFAULT_RESTORATION_TIER;
+    extern const std::string  S3_DEFAULT_STORAGE_CLASS;
+    extern const std::string  S3_STORAGE_CLASS_STANDARD;
+    extern const std::string  S3_STORAGE_CLASS_GLACIER;
+    extern const std::string  S3_STORAGE_CLASS_DEEP_ARCHIVE;
+    extern const std::string  S3_STORAGE_CLASS_GLACIER_IR;
+}
```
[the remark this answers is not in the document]
JustinKyleJames: They were already in that header.  Having him here is totally unnecessary.  This section has been removed.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2094,3 +2094,466 @@ def test_missing_keyfile(self):
...
+
+                if rc1 != 0:
+                    _, _, rc1  = self.user0.run_icommand("iget -f {file1} {file1_get}".format(**locals()))
+                if rc2 != 0:
+                    _, _, rc2  = self.user0.run_icommand("iget -f {file2} {file2_get}".format(**locals()))
+
+                if rc1 == 0 and rc2 == 0:
+                    break
+
+                time.sleep(20)
```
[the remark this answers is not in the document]
JustinKyleJames: Done

s3/s3_transport/unit_tests/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(unit_tests LANGUAGES C CXX)
 
 find_package(IRODS 4.3.0 EXACT REQUIRED)
 
-set(IRODS_EXTERNALS_FULLPATH_S3 "/opt/irods-externals/libs3e4197a5e-0")    # without verbosity
+set(IRODS_EXTERNALS_FULLPATH_S3 "/opt/irods-externals/libs3e4674774-0")    # without verbosity
```
trel, line 6: why do we have to redefine this here?   can't we inherit somehow from the main CMakeLists.txt?

@SwooshyCueb any suggestions?  or is this just how it works?

trel:
force pushing everything makes it hard to see what's changing.

s3/s3_transport/unit_tests/CMakeLists.txt
SwooshyCueb: Yeah this should not be explicitly defined at all. We have macros for that.

JustinKyleJames: Can someone point me to how this would be done?  Also, I need to be able to select the version for development purposes.

JustinKyleJames: I changed it to the following which matches what is done in the non-unit test cmake file.

```
include(IrodsExternals)
IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(S3 libs3e4674774-0)
string(REPLACE ";" ", " IRODS_PACKAGE_DEPENDENCIES_STRING "${IRODS_PACKAGE_DEPENDENCIES_LIST}")
```

SwooshyCueb: This is fine, however it is curious to me that it needs to be redefined in the first place

SwooshyCueb: Unless you are packaging the unit tests separately, the `IRODS_PACKAGE_DEPENDENCIES_STRING` is not needed.

packaging/resource_suite_s3_cache.py
```diff
@@ -40,7 +40,34 @@
...
+
+            # just write some random byte each 1024 chars
+            for i in range(0, to_write, 1024):
+                buffer[i] = current_char
+                current_char = random.randrange(256)
+            buffer[len(buffer)-1] = random.randrange(256)
+
+            out.write(bytearray(buffer))
+
+            bytes_written += to_write
```
alanking, line 70: If `truncate` is sufficiently arbitrary, we could use `make_file()` in `lib` with `contents='arbitrary'`. Otherwise, `contents='random'` will use `/dev/urandom` and `dd` to generate a file with random contents (might be slow when entropy is low). Thoughts?

s3/s3_operations.cpp
```diff
@@ -1144,6 +1146,56 @@ namespace irods_s3 {
...
+            if (0 == stat(cache_file.string().c_str(), &st)) {
+
+                _statbuf->st_size = st.st_size;
+
+            } else {
+
+                // Not using a cache file.  Read the max byte written from interprocess shmem.
+
+                std::string bucket_name;
+                std::string shmem_key = irods::experimental::io::s3_transport::constants::SHARED_MEMORY_KEY_PREFIX +
```
alanking, line 1179: `irods::experimental::io::s3_transport::constants::SHARED_MEMORY_KEY_PREFIX` -> `transport::constants::SHARED_MEMORY_KEY_PREFIX`

s3/s3_operations.cpp
```diff
@@ -1144,6 +1146,56 @@ namespace irods_s3 {
...
+                    std::to_string(std::hash<std::string>{}(resource_name + "/" + key));
+
+
+                irods::experimental::interprocess::shared_memory::named_shared_memory_object<transport::shared_data::multipart_shared_data>
```
alanking, line 1183: Consider using a type or namespace alias here for readability.

s3/s3_operations.cpp
```diff
@@ -1144,6 +1146,56 @@ namespace irods_s3 {
...
+                    shm_obj{shmem_key,
+                        180,  // shmem timeout
```
alanking, line 1185: A constant could be used here instead of the comment (or even a `#define`):
```c++
constexpr auto shmem_timeout = 180;
```

s3/s3_operations.cpp
```diff
@@ -416,6 +416,7 @@ namespace irods_s3 {
 
         if (oflag & O_APPEND) {
             mode |= ios_base::app;
+            mode &= ~ios_base::trunc;  // turn off trunc flag
```
alanking, line 419: Just to be clear, this particular translation is an issue only for S3, right? Meaning, if `trunc` is set in the openmode, the open of the S3 object - not the iRODS replica - is what fails. Just making sure I understand

s3/s3_transport/unit_tests/CMakeLists.txt
```diff
@@ -3,7 +3,10 @@ project(unit_tests LANGUAGES C CXX)
 
 find_package(IRODS 4.3.0 EXACT REQUIRED)
 
-set(IRODS_EXTERNALS_FULLPATH_S3 "/opt/irods-externals/libs3e4674774-0")    # without verbosity
+include(IrodsExternals)
+
+IRODS_MACRO_CHECK_DEPENDENCY_SET_FULLPATH_ADD_TO_IRODS_PACKAGE_DEPENDENCIES_LIST(S3 libs3e4674774-0)
+#string(REPLACE ";" ", " IRODS_PACKAGE_DEPENDENCIES_STRING "${IRODS_PACKAGE_DEPENDENCIES_LIST}")
```
alanking, line 9: Remove commented line?

alanking (commented):
Overall I think this is fine. Still have some suggestions

s3/s3_operations.cpp
JustinKyleJames: Would this mean holding off in 1978 until 4.3.1 and 4.2.12?

s3/s3_transport/src/s3_transport.cpp
```diff
@@ -134,18 +148,31 @@ namespace irods::experimental::io::s3_transport
...
 
-        const auto xml = fmt::format("<RestoreRequest>\n "
-                                     "  <Days>{}</Days>\n"
-                                     "  <GlacierJobParameters>\n"
-                                     "    <Tier>{}</Tier>\n"
-                                     "  </GlacierJobParameters>]n"
-                                     "</RestoreRequest>\n",
-                                     restoration_days,
-                                     restoration_tier);
+        // restoration_tier is not valid for DEEP_ARCHIVE
```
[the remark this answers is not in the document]
JustinKyleJames: done

packaging/resource_suite_s3_cache.py
JustinKyleJames: I am not sure I understand.  I did not want to use up entropy which is why I used a library that only uses pseudo-random numbers.  This is also why I didn't use `contents='random'`.

Using `contents='arbitrary'` simply truncates a file which is not sufficient for my needs.

s3/s3_operations.cpp
JustinKyleJames: done

s3/s3_operations.cpp
JustinKyleJames: done

s3/s3_operations.cpp
JustinKyleJames: This is an issue if we are using a cache file.  I don't remember exactly but I believe O_APPEND and O_TRUNC are allowed in POSIX but the combination of std::ios_base::append and std::ios_base::trunc are not allowed in streams.

s3/s3_transport/unit_tests/CMakeLists.txt
JustinKyleJames: done

s3/s3_operations.cpp
JustinKyleJames: done

packaging/resource_suite_s3_cache.py
JustinKyleJames: Doing some testing, it looks like this might actually use some entropy.  I'm not sure why as hardcoding the seed and then running through the sequence I get the same results each time.

Creating a 1G file, the entropy goes down slightly but not significantly.  Either way, I can rewrite this.

JustinKyleJames: We discussed this.   The random library with a hardcoded seed does not use up entropy which is what we want.

Using `contents='arbitrary'` simply truncates a file which is not sufficient for my needs and `contents='random'` is more than I need as I don't need true randomness.

What I need is to make sure that each 1k bytes is unique.  This should be adequate to catch an algorithmic issue within the plugin while sending or receiving data.

s3/s3_operations.cpp
JustinKyleJames: In that case, I will rollback the change for 1978 for now.

packaging/resource_suite_s3_cache.py
alanking: Yep, sounds good. We could add a new option like `contents='pseudorandom'` :p

trel:
> Note: The libs3 version will need to be updated before we merge.

is this still true?   or are we good on libs3?

JustinKyleJames:
Ping to take another look after rebase on issue 2082.

packaging/resource_suite_s3_cache.py
```diff
@@ -608,3 +635,237 @@ def test_multiple_s3_endpoints_replication_issue1858(self):
...
+                              secret_key=self.aws_secret_access_key,
+                              http_client=httpClient,
+                              region=self.s3region)
+        else:
+            s3_client = Minio(self.s3endPoint,
+                              access_key=self.aws_access_key_id,
+                              secret_key=self.aws_secret_access_key,
+                              http_client=httpClient,
+                              region=self.s3region,
+                              secure=False)
```
korydraughn, line 757: These can be collapsed where `secure=(self.proto == 'HTTPS')`.

packaging/resource_suite_s3_cache.py
```diff
@@ -608,3 +635,237 @@ def test_multiple_s3_endpoints_replication_issue1858(self):
...
+            admin_session.assert_icommand("iadmin rmresc cacheResc")
+            admin_session.assert_icommand("iadmin rmresc demoResc")
+            admin_session.assert_icommand("iadmin modresc origResc name demoResc", 'STDOUT_SINGLELINE', 'rename', input='yes\n')
+
+        shutil.rmtree(IrodsConfig().irods_directory + "/cacheRescVault", ignore_errors=True)
+
+    def get_resource_context(self, resc_name):
+
+        resource_context = self.admin.run_icommand('iquest "%s" "SELECT RESC_CONTEXT where RESC_NAME = \'{resc_name}\'"'.format(**locals()))[0].strip()
+        return resource_context
```
korydraughn, line 778: The body could be a one liner - `return self.admin.run_icommand( ...`

packaging/resource_suite_s3_nocache.py
```diff
@@ -2094,3 +2094,460 @@ def test_missing_keyfile(self):
...
+        hostname = lib.get_hostname()
+
+        # read aws keys
+        self.read_aws_keys()
+
+        # set up s3 bucket
+        if self.proto == 'HTTPS':
+            s3_client = Minio(self.s3endPoint, access_key=self.aws_access_key_id, secret_key=self.aws_secret_access_key, region = self.s3region)
+        else:
+            s3_client = Minio(self.s3endPoint, access_key=self.aws_access_key_id, secret_key=self.aws_secret_access_key, region = self.s3region, secure=False)
```
korydraughn, line 2134: Can be a one liner. See earlier comment.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2094,3 +2094,460 @@ def test_missing_keyfile(self):
...
+            admin_session.run_icommand(['iadmin', 'rmresc', self.anotherresc])
+            admin_session.assert_icommand("iadmin rmresc demoResc")
+            admin_session.assert_icommand("iadmin modresc origResc name demoResc", 'STDOUT_SINGLELINE', 'rename', input='yes\n')
+            print("run_resource_teardown - END")
+
+        # delete s3 bucket
+        if self.proto == 'HTTPS':
+            s3_client = Minio(self.s3endPoint, access_key=self.aws_access_key_id, secret_key=self.aws_secret_access_key, region=self.s3region)
+        else:
+            s3_client = Minio(self.s3endPoint, access_key=self.aws_access_key_id, secret_key=self.aws_secret_access_key, region=self.s3region, secure=False)
```
korydraughn, line 2209: Same as earlier comment.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2094,3 +2094,460 @@ def test_missing_keyfile(self):
...
+        finally:
+
+            # cleanup
+
+            self.user0.assert_icommand("irm -f {file1}".format(**locals()), 'EMPTY')
+            self.user0.assert_icommand("irm -f {file2}".format(**locals()), 'EMPTY')
+            self.admin.assert_icommand("iadmin rmresc s3resc1", 'EMPTY')
+
+            if os.path.exists(file1):
+                os.unlink(file1)
```
korydraughn, line 2425: This can be put in a function and used in many of the tests.
```python
lib.remove_if_exists(file)
```
The function could exist n this repo and lib.py.

korydraughn (commented):
Looks good overall.

Most of my comments apply to multiple tests.

packaging/resource_suite_s3_cache.py
JustinKyleJames: Done for all.

packaging/resource_suite_s3_cache.py
JustinKyleJames: Done for all

packaging/resource_suite_s3_nocache.py
JustinKyleJames: Done for all.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: I added a local lib name s3plugin_lib.py for this.

Happy to add it to the main lib later but that would be a pull request for irods.

korydraughn (commented):
The changes look good.

Consider adding those python helper functions to irods/irods. It's not urgent, but those would be good to have in lib.py at some point.

alanking:
We can create an issue and/or open a PR there as well as keep the methods here so that we can go ahead and get this in. Thoughts?

korydraughn:
I agree. Let's get this in and we can port the python helper functions to lib.py later.

alanking:
Okey dokey. Please squash and leave off #'s, @JustinKyleJames

s3/libirods_s3.cpp
```diff
@@ -2169,8 +2169,8 @@ irods:: error s3StartOperation(irods::plugin_property_map& _prop_map)
             return SUCCESS();
         }
 
-        // Look up the rodsServerHost_t for this server and set this as the RESOURCE_HOST of the resource.
-        // The rodsServerHost_t entry is shared throughout the code (not just used for this
+        // Look up the rodsHostAddr_t for this server and set this as the RESOURCE_HOST of the resource.
+        // Issue 2082: The rodsHostAddr_t entry is shared throughout the code (not just used for this
```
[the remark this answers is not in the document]
alanking: Oops. By "this commit" I meant [50fea9e](https://github.com/irods/irods_resource_plugin_s3/pull/2077/commits/50fea9e733ce70260246d0e05d846f08acea7cfa)

packaging/resource_suite_s3_nocache.py
alanking: I think that #2088 addresses this. Please mark this as resolved if true, @JustinKyleJames or @korydraughn. Thanks!

s3/libirods_s3.cpp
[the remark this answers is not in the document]
JustinKyleJames: I cleaned that up.

alanking (approved):

packaging/resource_suite_s3_nocache.py
JustinKyleJames: I added an issue to have these moved over to lib.py.  Resolving this conversation.

korydraughn (approved):

alanking:
I think we are ready for #s

JustinKyleJames:
Pounds added.

PR #2096, opened 2023-02-09
[irods/irods6502] enable test case with archive/glacier (4-2-stable)

JustinKyleJames:


packaging/resource_suite_s3_cache.py
```diff
@@ -649,7 +649,7 @@ def setUp(self):
         if hasattr(self, 'static_bucket_name'):
             self.s3bucketname = self.static_bucket_name
         else:
-            distro_str = '{}-{}'.format(distro.id(), distro.version()).replace(' ', '').replace('.', '')
+            distro_str = ''.join(platform.linux_distribution()[:2]).replace(' ','').replace('.', '')
```
alanking, line 652: Is `distro` not working here? I thought this was the preferred module these days...

alanking (commented):
Just one question. The rest looks good as soon as we're ready to merge irods/irods#6899

packaging/resource_suite_s3_cache.py
korydraughn: I don't think we moved to `distro` for 4-2-stable, so I think this is okay?

Willing to be persuaded otherwise.

alanking: Oh, I see. This is only in the tests which were being skipped until now, and so it only now presents a problem because `distro` isn't being used/supported in this branch. As long as we don't do this in the cherry-pick to main, then this is fine.

alanking (approved):
I think this is good to go.

korydraughn:
@JustinKyleJames Have you seen the these tests pass?

JustinKyleJames:
> @JustinKyleJames Have you seen the these tests pass?

Yes, these have passed for me.

korydraughn (approved):
Please add the pound.

packaging/resource_suite_s3_cache.py
JustinKyleJames: This failed which is why I made the change.

JustinKyleJames:
> Please add the pound.

Done

alanking:
Are the changes in irods/irods#6899 required for this test to pass? If so, I'm going to wait and merge these at the same time unless anybody objects. If it passes with things just as they are now, then I will go ahead and merge this without waiting for the other PR.

Issue #2095, opened 2023-01-30
`irods` service account user has no permissions on the S3 keypair file (test hook)


alanking:
- [x] main
- [x] 4-2-stable

PR #2094, opened 2023-01-30
Add tests for data objects stuck in locked status, prevent nullptr dereference (main)

alanking:
Addresses irods/irods#6154
Addresses irods/irods#6479
Addresses irods/irods#6880

Companion PR: https://github.com/irods/irods/pull/6884

Running tests now.

Issue #2092, opened 2023-01-25
Build problems: build hook externals and libs3 dependency


alanking:
- [x] main
- [x] 4-2-stable
---
I have been experiencing some problems with building packages for the main branch, and it stems from 3 things:

1. The libs3 dependency declared in CMakeLists needs to be bumped to https://github.com/irods/libs3/commit/c0e278d280cde08fb588d82dcb125198852f2bc9
2. The build hook needs to install all relevant externals because installing custom externals from a local directory will not install everything that needs to be installed (i.e. revert external removals in abd496db7bc57a51a9b37c81a3a9c0c69470a337)
3. The libs3 dependency in the build hook (and test hook?) needs to be bumped to match the one in CMakeLists

alanking:
I think we need this for 4-2-stable as well... Re-opening until proven otherwise.

PR #2097, opened 2023-02-10
Add tests for data objects stuck in locked status, prevent nullptr dereference (4-2-stable)

alanking:
Addresses #2092
Addresses #2095
Addresses https://github.com/irods/irods/issues/6154
Addresses https://github.com/irods/irods/issues/6479
Addresses https://github.com/irods/irods/issues/6880

Companion PR: https://github.com/irods/irods/pull/6905
Companion PR: https://github.com/irods/externals/pull/181

Cherry-picked from https://github.com/irods/irods_resource_plugin_s3/pull/2094
Cherry-picked from https://github.com/irods/irods_resource_plugin_s3/pull/2093

Leaving in draft for now because building the plugin doesn't quite work. Might need to do something in irods/externals first...


korydraughn:
Should we wait to review this?

alanking:
This can be reviewed now as it is nearly identical to #2094, just reserving the right to do some unannounced force-pushing :p

korydraughn:
Will review as soon as i get an opening.

alanking:
https://github.com/irods/externals/pull/181 is required along with the current set of changes in order to build the plugin. Will start running tests soon. Marking as ready for review since I do not expect to make any more dramatic changes outside of review comments.

alanking:
When I tried to run tests for this, it seemed to be infinitely stuck and a lot of error messages were coming out of MinIO. Will need to investigate. Everything leading up to running the tests (including running the test hook) seems fine now.

alanking:
A couple of things happened, preventing the tests from passing here:

1. Running two MinIO servers on the same host seems to be causing some problems when running tests in the testing environment. Not sure why this hasn't cropped up before, but it is causing problems now. I ran a single MinIO server and tests started running normally. As such, I created this issue to address these problems: #2101
2. A bug was discovered while trying to run my new tests in that an error is not being returned when copying data to S3 fails during sync-to-archive. @JustinKyleJames found the fix, which I've replicated in the latest commit. I am happy to drop that commit so he can get proper credit, but this set of changes causes tests to pass.

So... everything passes now, yay

korydraughn:
> This can be reviewed now as it is nearly identical to #2094, just reserving the right to do some unannounced force-pushing :p

Is this PR still **nearly** identical to #2094? Is there anything we need to look at specifically since the new commits?

alanking:
The only things that should be different at this point are the changes to the test hook to chown the keypair file (which is different because of python2) and the latest commit which addresses #2100 (which needs its commit message updated and/or @JustinKyleJames should make that change in a separate PR). So, it's still nearly identical save for those two things.

korydraughn (commented):
Changes look good. Will await outcome with @JustinKyleJames regarding commit ownership before approving this PR.

JustinKyleJames:
> Changes look good. Will await outcome with @JustinKyleJames regarding commit ownership before approving this PR.

Commit looks good and I think I am fine with the ownership being as is.

alanking:
Thanks! Squashed and added coauthorship to that last commit.

korydraughn (approved):
Add the pounds!

korydraughn:
Merge when ready.

PR #2098, opened 2023-02-13
[#2092] Shave the libs3 sha (main)

alanking:
Addresses #2092

Depends on irods/externals#182


korydraughn (approved):
It matches the value in externals.

Ship it when ready.

PR #2105, opened 2023-04-20
[2102] Get the number of threads from destination dataObjInp (4-2-stable)

JustinKyleJames:
Added test case

### DIFF
--- a/packaging/resource_suite_s3_nocache.py
+++ b/packaging/resource_suite_s3_nocache.py
@@ -2198,6 +2198,84 @@ OUTPUT ruleExecOut
                     'DATA_REPL_STATUS', '0'])
             self.user0.run_icommand(['irm', '-f', logical_path])
 
+    # Replication to an S3 resource failed for file sizes between [4MB, 32MB] due to not knowing the correct
+    # number of write threads being used.  Test the following conditions:
+    # 1) files just inside this range
+    # 2) files just outside the range
+    # 3) a couple of files in the middle of the range (8MB and 16MB)
+    # 4) a small file
+    # 5) a file one byte less than 32MB just in case
+    def test_s3_in_replication_node__issue_2102(self):
+        resc_prefix = 'issue_2101'
+        repl_resource_name = '%s_repl' % resc_prefix
+        s3_resc1_resource_name = '%s_s3_resc1' % resc_prefix
+        s3_resc2_resource_name = '%s_s3_resc2' % resc_prefix
+        s3_resc3_resource_name = '%s_s3_resc3' % resc_prefix
+        filename = '%s_file' % resc_prefix
+        hostname = lib.get_hostname()
+
+        try:
+
+            # set up resource tree
+            self.admin.assert_icommand("iadmin mkresc %s replication" % repl_resource_name, 'STDOUT_SINGLELINE', "Creating")
+            self.admin.assert_icommand("iadmin mkresc %s s3 %s:/%s/%s %s" %
+                                   (s3_resc1_resource_name, hostname, self.s3bucketname, s3_resc1_resource_name, self.s3_context), 'STDOUT_SINGLELINE', "Creating")
+            self.admin.assert_icommand("iadmin mkresc %s s3 %s:/%s/%s %s" %
+                                   (s3_resc2_resource_name, hostname, self.s3bucketname, s3_resc2_resource_name, self.s3_context), 'STDOUT_SINGLELINE', "Creating")
+            self.admin.assert_icommand("iadmin mkresc %s s3 %s:/%s/%s %s" %
+                                   (s3_resc3_resource_name, hostname, self.s3bucketname, s3_resc3_resource_name, self.s3_context), 'STDOUT_SINGLELINE', "Creating")
+            self.admin.assert_icommand("iadmin addchildtoresc %s %s" % (repl_resource_name, s3_resc1_resource_name))
+            self.admin.assert_icommand("iadmin addchildtoresc %s %s" % (repl_resource_name, s3_resc2_resource_name))
+            self.admin.assert_icommand("iadmin addchildtoresc %s %s" % (repl_resource_name, s3_resc3_resource_name))
+
+            file_sizes = [
+                    100,
+                    4 * 1024 * 1024 - 1,
+                    4 * 1024 * 1024,
+                    8 * 1024 * 1024,
+                    16 * 1024 * 1024,
+                    32 * 1024 * 1024 - 1,
+                    32 * 1024 * 1024,
+                    32 * 1024 * 1024 + 1,
+                    ];
+
+            for file_size in file_sizes:
+                # create the file
+                lib.make_arbitrary_file(filename, file_size)
+
+                # put the file to repl_resource
+                self.user1.assert_icommand("iput -R %s %s" % (repl_resource_name, filename))
+
+                # debug
+                self.user1.assert_icommand("ils -l %s" % filename, 'STDOUT_SINGLELINE', filename)
+
+                # get each replica
+                self.user1.assert_icommand("iget -f -n 0 %s %s_get0" % (filename, filename))
+                self.user1.assert_icommand("diff %s %s_get0" % (filename, filename), 'EMPTY')
+                self.user1.assert_icommand("iget -f -n 1 %s %s_get1" % (filename, filename))
+                self.user1.assert_icommand("diff %s %s_get1" % (filename, filename), 'EMPTY')
+                self.user1.assert_icommand("iget -f -n 2 %s %s_get2" % (filename, filename))
+                self.user1.assert_icommand("diff %s %s_get2" % (filename, filename), 'EMPTY')
+
+                # cleanup before next loop
+                s3plugin_lib.remove_if_exists(filename)
+                s3plugin_lib.remove_if_exists("%s_get0" % filename)
+                s3plugin_lib.remove_if_exists("%s_get1" % filename)
+                s3plugin_lib.remove_if_exists("%s_get2" % filename)
+                self.user1.assert_icommand("irm -f %s" % (filename))
+
+        finally:
+            # clean up the file that was put in case it wasn't cleaned up
+            self.user1.assert_icommand("irm -f %s" % (filename))
+
+            # clean up resource tree
+            self.admin.assert_icommand("iadmin rmchildfromresc %s %s" % (repl_resource_name, s3_resc1_resource_name))
+            self.admin.assert_icommand("iadmin rmchildfromresc %s %s" % (repl_resource_name, s3_resc2_resource_name))
+            self.admin.assert_icommand("iadmin rmchildfromresc %s %s" % (repl_resource_name, s3_resc3_resource_name))
+            self.admin.assert_icommand("iadmin rmresc %s" % repl_resource_name)
+            self.admin.assert_icommand("iadmin rmresc %s" % s3_resc1_resource_name)
+            self.admin.assert_icommand("iadmin rmresc %s" % s3_resc2_resource_name)
+            self.admin.assert_icommand("iadmin rmresc %s" % s3_resc3_resource_name)
 
 class Test_S3_NoCache_Glacier_Base(session.make_sessions_mixin([('otherrods', 'rods')], [('alice', 'apass'), ('bobby', 'bpass')])):
 
--- a/s3/s3_operations.cpp
+++ b/s3/s3_operations.cpp
@@ -156,14 +156,15 @@ namespace irods_s3 {
 
         // default - object need not exist
         return false;
-    } // end operation_requires_that_object_exists 
+    } // end operation_requires_that_object_exists
 
     irods::error s3_file_stat_operation_with_flag_for_retry_on_not_found(irods::plugin_context& _ctx,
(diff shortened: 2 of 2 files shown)


trel:
ah, and consolidated to the single for loop.

looking good.

s3/s3_operations.cpp
```diff
@@ -245,19 +245,44 @@ namespace irods_s3 {
...
 
             for (int i = 0; i < NUM_L1_DESC; ++i) {
-
-                if (L1desc[i].inuseFlag && L1desc[i].dataObjInp && L1desc[i].dataObjInfo
-                       && L1desc[i].dataObjInp->objPath == file_obj->logical_path()
-                       && L1desc[i].dataObjInp->oprType == REPLICATE_SRC ) {
-
+                // get the data size from source dataObjInfo
+                if (!found_data_size && L1desc[i].inuseFlag && L1desc[i].dataObjInp && L1desc[i].dataObjInfo &&
+                    L1desc[i].dataObjInp->objPath == file_obj->logical_path() &&
```
alanking, line 257: Consider doing check at the top of the loop which will `continue` if true:
```c++
if (L1desc[i].inuseFlag &&
    L1desc[i].dataObjInp &&
    L1desc[i].dataObjInfo &&
    L1desc[i].dataObjInp->objPath == file_obj->logical_path())
{
    continue;
}
```
This would simplify the two checks on 256 and 268.

s3/s3_operations.cpp
```diff
@@ -245,19 +245,44 @@ namespace irods_s3 {
...
+                            __LINE__,
+                            __FUNCTION__,
+                            thread_id,
+                            number_of_threads);
+
+                    found_number_of_threads = true;
+                }
+
+                // once we have both pieces of information break out of for loop
+                if (found_data_size && found_number_of_threads) {
```
korydraughn, line 285: Are these allowed to be set on different iterations of the loop?
Should they be tied to the same L1desc?

packaging/resource_suite_s3_nocache.py
```diff
@@ -2198,6 +2198,73 @@ def test_iput_and_icp_with_invalid_secret_key__issue_6154(self):
...
+
+                # cleanup before next loop
+                s3plugin_lib.remove_if_exists(filename)
+                s3plugin_lib.remove_if_exists("%s_get0" % filename)
+                s3plugin_lib.remove_if_exists("%s_get1" % filename)
+                s3plugin_lib.remove_if_exists("%s_get2" % filename)
+                self.user1.assert_icommand("irm -f %s" % (filename))
+
+        finally:
+            # clean up resource tree
```
alanking, line 2260: Might want to add another `irm -f` down here because it might not be able to clean up the resources if the test fails before the `irm -f` above.

alanking (commented):
Overall looks real good. Just had a couple of thoughts

s3/s3_operations.cpp
```diff
@@ -245,19 +245,44 @@ namespace irods_s3 {
...
             for (int i = 0; i < NUM_L1_DESC; ++i) {
-
-                if (L1desc[i].inuseFlag && L1desc[i].dataObjInp && L1desc[i].dataObjInfo
-                       && L1desc[i].dataObjInp->objPath == file_obj->logical_path()
-                       && L1desc[i].dataObjInp->oprType == REPLICATE_SRC ) {
-
+                // get the data size from source dataObjInfo
+                if (!found_data_size && L1desc[i].inuseFlag && L1desc[i].dataObjInp && L1desc[i].dataObjInfo &&
+                    L1desc[i].dataObjInp->objPath == file_obj->logical_path() &&
+                    L1desc[i].dataObjInp->oprType == REPLICATE_SRC)
```
korydraughn, line 258: Consider using references and/or pointers to the members of interest. For example:
```c++
const auto& l1d = L1desc[i];
const auto* dobj_input = l1d.dataObjInp;
const auto* dobj_info = l1d.dataObjInfo;
```

s3/s3_operations.cpp
```diff
@@ -245,19 +245,44 @@ namespace irods_s3 {
             }
         }
 
-
-        // if this is a replication and we're the destination, get the data size from the source dataObjInfo
+        // special treatment for replication - get data_size from source and
+        // number_of_threads from destination
```
korydraughn, line 249: Consider tweaking this comment so that it explains **why** we do this.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2198,6 +2198,73 @@ def test_iput_and_icp_with_invalid_secret_key__issue_6154(self):
                     'DATA_REPL_STATUS', '0'])
             self.user0.run_icommand(['irm', '-f', logical_path])
 
+    def test_s3_in_replication_node__issue_2102(self):
```
korydraughn, line 2201: Consider using more words so that the intent of the test is not ambiguous.

s3/s3_operations.cpp
JustinKyleJames: No, because one bit of information is taken from the replication source and the other from the replication destination.

s3/s3_operations.cpp
JustinKyleJames: done

s3/s3_operations.cpp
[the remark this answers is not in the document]
JustinKyleJames: done except for the comment left above about dataObjInfo not being needed in the second loop.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: done

s3/s3_operations.cpp
JustinKyleJames: done

packaging/resource_suite_s3_nocache.py
JustinKyleJames: done

s3/s3_operations.cpp
korydraughn: I think the loop allows for that possibility.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2198,6 +2198,73 @@ def test_iput_and_icp_with_invalid_secret_key__issue_6154(self):
...
+                    32 * 1024 * 1024 - 1,
+                    32 * 1024 * 1024,
+                    ];
+
+            for file_size in file_sizes:
+                # create the file
+                lib.make_arbitrary_file(filename, file_size)
+
+                # put the file to repl_resource
+                self.user1.assert_icommand("iput -R %s %s" % (repl_resource_name, filename))
```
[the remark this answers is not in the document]
JustinKyleJames: I made an attempt to do this.  It is in the latest commit:

Some notes:

1. I copied the entire contents of the file into a bytes buf.  If you do multiple istream with append it is really testing something different. 
2. I had to use a local version of make_arbitrary_file updated so that the bytes were all within 0-256.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2198,6 +2198,98 @@ def test_iput_and_icp_with_invalid_secret_key__issue_6154(self):
...
+                    s3plugin_lib.remove_if_exists(filename)
+                    s3plugin_lib.remove_if_exists("%s_get0" % filename)
+                    s3plugin_lib.remove_if_exists("%s_get1" % filename)
+                    s3plugin_lib.remove_if_exists("%s_get2" % filename)
+                    self.user1.assert_icommand("irm -f %s" % (filename))
+
+        finally:
+            # clean up the file that was put in case it wasn't cleaned up
+            try:
+                self.user1.assert_icommand("irm -f %s" % (filename))
```
alanking, line 2281: Is there a reason this is wrapped with the try-except? `irm -f` will not fail if the object does not exist, and if it fails for some other reason, then removing the resources below will fail as well.

packaging/s3plugin_lib.py
```diff
@@ -21,13 +21,13 @@ def make_arbitrary_file(f_name, f_size, buffer_size=32*1024*1024):
             else:
                 to_write = buffer_size
 
-            current_char = random.randrange(256)
+            current_char = random.randrange(128)
```
alanking, line 24: Is the 256 -> 128 intentional?

s3/s3_operations.cpp
```diff
@@ -245,20 +245,51 @@ namespace irods_s3 {
...
         if (oprType == REPLICATE_DEST) {
+            bool found_data_size = false;
+            bool found_number_of_threads = false;
 
             for (int i = 0; i < NUM_L1_DESC; ++i) {
+                const auto& l1d = L1desc[i];
+                const auto* dobj_input = l1d.dataObjInp;
+                const auto* dobj_info = l1d.dataObjInfo;
+
+                if (l1d.inuseFlag && dobj_input && dobj_input->objPath == file_obj->logical_path()) {
```
alanking, line 260: If this was inverted, you could use `continue;` and save an indent.

Feel free to ignore this!

alanking (commented):
I'm not sure that I fully understand the value of the `istream` test since my understanding of the failure is that the replication resource fileModified operation was not working properly in this case, not the initial operation which creates the object. But if we feel that it is useful, then I'm all for more tests. :)

s3/s3_operations.cpp
```diff
@@ -245,20 +245,51 @@ namespace irods_s3 {
...
+                    }
 
-                    data_size = L1desc[i].dataObjInfo->dataSize;
-                    rodsLog(developer_messages_log_level, "%s:%d (%s) [[%lu]] repl to s3 destination.  setting data_size to %zd\n",
-                            __FILE__, __LINE__, __FUNCTION__, thread_id, data_size);
-                    break;
+                    // once we have both pieces of information break out of for loop
+                    if (found_data_size && found_number_of_threads) {
+                        break;
+                    }
```
korydraughn, line 292: If these are required to be true on the same iteration, then we need to add a check that enforces that requirement.

It can be as simple as checking that they are always equal just before or after this if-block.

korydraughn:
> I'm not sure that I fully understand the value of the `istream` test since my understanding of the failure is that the replication resource fileModified operation was not working properly in this case, not the initial operation which creates the object. But if we feel that it is useful, then I'm all for more tests. :)

If you and @JustinKyleJames agree that including `istream` in the test isn't necessary, then it is fine to remove it.

trel:
yeah, this is about the replication, not the initial incoming data.   don't think `istream` adds anything to this scenario.

packaging/s3plugin_lib.py
JustinKyleJames: Yes, I was having issues with string encoding otherwise.

JustinKyleJames:
> I'm not sure that I fully understand the value of the `istream` test since my understanding of the failure is that the replication resource fileModified operation was not working properly in this case, not the initial operation which creates the object. But if we feel that it is useful, then I'm all for more tests. :)

I agree.  I added it because I was asked to.  Thoughts?

packaging/s3plugin_lib.py
JustinKyleJames: Since we are removing the istream.  I'll change this back.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: I didn't realize that it would not fail if the object does not exist.  I'll remove the try/except.

s3/s3_operations.cpp
JustinKyleJames: I changed the method to return an irods::error() in this case.  Please take a look at it.

s3/s3_operations.cpp
JustinKyleJames: Done

s3/s3_operations.cpp
JustinKyleJames: Actually these are not required to be true on the same iteration.  In fact they will never be true on the same iteration.

I would expect them to be true at the end but checking for these both be be true at the end may have broken a test which means that something didn't behave the way I expected.  I am looking into it now.

JustinKyleJames: It appears that on an irepl -r there is not a replication source in the `L1desc` table.  Therefore, I can't guarantee that the `found_data_size` will be true.

This is acceptable because if the data size is not known it will default to using a cache file.  I kept in the check to make sure `found_number_of_threads` was true but took out the one for `found_data_size`.

I am running it through some tests and if it passes I will commit the latest change.

korydraughn: Ok. Will wait to see what you find out.

JustinKyleJames: Running one suite of cacheless tests passed.  This should be good to go. 

Note that in the scenario I referenced, the destination L1desc entry does have a srcL1descIdx but the actual entry at that index is not valid or is no longer valid.

korydraughn:
Does this PR pass the tests?

Anything else left to do for this PR?

JustinKyleJames:
> Does this PR pass the tests?
> 
> Anything else left to do for this PR?

This passes test.  There is nothing left to do for it.

korydraughn:
Please squash to taste and leave off pounds.

korydraughn (approved):
Add the pound when ready.

JustinKyleJames:
> Add the pound when ready.

Done

PR #2107, opened 2023-04-26
[2102] Get the number of threads from destination dataObjInp

JustinKyleJames:
This is the same as the one for 4-2-stable except for:

1. The rodsLog() statements have been replaced by the logger.
2. I ran clang format.  (I did not keep the spaces to tabs conversion.)

The tests passed.

### DIFF
--- a/s3/s3_operations.cpp
+++ b/s3/s3_operations.cpp
@@ -163,11 +163,13 @@ namespace irods_s3 {
             struct stat* _statbuf, bool retry_on_not_found );
 
     // determines the data size and number of threads, stores them, and returns them
-    void get_number_of_threads_data_size_and_opr_type(irods::plugin_context& _ctx,
-        int& number_of_threads, std::int64_t& data_size, int& oprType, bool query_metadata = true) {
-
+    auto get_number_of_threads_data_size_and_opr_type(irods::plugin_context& _ctx,
+                                                      int& number_of_threads,
+                                                      int64_t& data_size,
+                                                      int& oprType,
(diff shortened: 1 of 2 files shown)


korydraughn (approved):
Looks good to me.

Add pound when ready.

s3/s3_operations.cpp
```diff
@@ -487,9 +523,12 @@ namespace irods_s3 {
...
         }
 
-        get_number_of_threads_data_size_and_opr_type(_ctx, number_of_threads, data_size, oprType);
+		ret = get_number_of_threads_data_size_and_opr_type(_ctx, number_of_threads, data_size, oprType);
+		if (!ret.ok()) {
+			return std::make_tuple(PASS(ret), data.dstream_ptr, data.s3_transport_ptr);
+		}
 
-        logger::debug("{}:{} ({}) [[{}]] oprType set to {}", __FILE__, __LINE__, __FUNCTION__, thread_id, oprType);
+		logger::debug("{}:{} ({}) [[{}]] oprType set to {}", __FILE__, __LINE__, __FUNCTION__, thread_id, oprType);
```
alanking, line 531: See some tabs down here in this file

JustinKyleJames: I fixed that and a couple of other places.

There are still a couple of code segments that have tabs from previous pull requests/formats.  I can fix that later but don't want to confuse people by making those changes here.

JustinKyleJames:
I fixed the tab/space issue and pounded it.

alanking (approved):

PR #2116, opened 2023-05-11
Issue 2114 (main)

JustinKyleJames:
This is a change to reset the thread pool when uploading a large file from a local cache file.  

Notes:

1. The commit for issue 2115 is simply moving some tests to a new class so that they don't get executed in every scenario.  The contents of those tests are unaltered.
2. Before release there will be a new version of libs3 and the CMakeLists.txt for this will need to be updated.  Since I don't know the final hash for that libs3 I can't make the update now.

### DIFF
--- a/packaging/resource_suite_s3_nocache.py
+++ b/packaging/resource_suite_s3_nocache.py
@@ -2325,6 +2325,47 @@ OUTPUT ruleExecOut
             self.admin.assert_icommand("iadmin rmresc %s" % s3_resc2_resource_name)
             self.admin.assert_icommand("iadmin rmresc %s" % s3_resc3_resource_name)
 
+# The tests in this class take a long time and do not need to run in every different test suite
+class Test_S3_NoCache_Large_File_Tests_Base(Test_S3_NoCache_Base):
+
+    @unittest.skipIf(psutil.disk_usage('/').free < 2 * (11*1024*1024*1024 + 2), "not enough free space for two 11 GiB files")
+    def test_upload_very_large_file_after_redirect_issues_2104_2114(self):
+
+        # in non-topology HOSTNAME_3 is just local host so it really doesn't test detached mode
+        # but in topology it will
+
+        file1 = "f1"
+        retrieved_file = "f1.get"
+        resource_host = test.settings.HOSTNAME_3
+        resource_name = "s3_resc_on_host3"
+
+        # create an S3 resource
+        # with S3_MPU_THREADS = 2 when testing in topology:
+        # - There will be a redirect and cache will be used.
+        # - There will only be 2 simultaneous part uploads but will require a third part to follow (issue 2114)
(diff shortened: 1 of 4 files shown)


s3/s3_transport/include/s3_transport.hpp
```diff
@@ -1081,13 +1081,14 @@ namespace irods::experimental::io::s3_transport
...
-                irods::thread_pool cache_flush_threads{static_cast<int>(config_.number_of_cache_transfer_threads)};
-
                 unsigned int part_number = 1;
 
                 std::int64_t part_size_all_but_last_part = cache_file_size / number_of_parts;
+
                 while (part_number <= number_of_parts) {
 
+                    irods::thread_pool cache_flush_threads{static_cast<int>(config_.number_of_cache_transfer_threads)};
+
```
korydraughn, line 1091: This seems costly.

How often do you expect a new thread pool to be instantiated?

JustinKyleJames: Rarely.  For example, with the defaults of 10 threads and 5 GB maximum part size, for a 100 GB file it would get re-instantiated one time (2 total instantiations).  Whatever time it takes to re-instantiate it would only be a fraction compared to the time to actually send the bytes.

korydraughn: I feel much better about the change now.

Please mark this conversation as resolved.

korydraughn (commented):
Changes look good.

Do the tests pass?

JustinKyleJames:
> Changes look good.
> 
> Do the tests pass?

Yes, they did.

korydraughn (approved):
Let's wait for one more approval before adding the pounds.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2325,6 +2177,195 @@ def test_s3_in_replication_node__issue_2102(self):
...
+
+        try:
+            # create an 11 GB file 
+            lib.make_file(file1, 11*1024*1024*1024)
+
+            # put the 11 GB file 
+            self.user1.assert_icommand("iput %s" % file1)  # iput
+            self.user1.assert_icommand("iget %s %s" % (file1, retrieved_file))
+
+            assert filecmp.cmp(file1, retrieved_file)  # confirm retrieved is correct
```
alanking, line 2358: Consider using `self.assertTrue` instead of naked `assert`.

alanking (approved):
One minor/optional thing. Up to you

packaging/resource_suite_s3_nocache.py

alanking (commented):
Nice. Please squash and add #s

korydraughn (approved):

JustinKyleJames:
> Nice. Please squash and add #s

Done

PR #2125, opened 2023-07-05
[2122] Reset global vars after last file closes. (4-2-stable)

JustinKyleJames:
Update S3 repl test to use checksum.

Do not run S3 repl test when in detached mode as we are using the same bucket for all resources thus the object keys conflict.

### DIFF
--- a/packaging/test_irods_resource_plugin_s3.py
+++ b/packaging/test_irods_resource_plugin_s3.py
@@ -150,6 +150,9 @@ class Test_S3_NoCache_Decoupled(Test_S3_NoCache_Base, unittest.TestCase):
     def test_put_get_large_file_in_repl_node(self):
         pass
 
+    @unittest.skip('test does not work in decoupled because we are using same bucket for multiple resources')
+    def test_s3_in_replication_node__issues_2102_2122(self):
+        pass
 
 class Test_S3_NoCache_Glacier(Test_S3_NoCache_Glacier_Base, unittest.TestCase):
 
--- a/s3/s3_operations.cpp
+++ b/s3/s3_operations.cpp
@@ -998,6 +998,20 @@ namespace irods_s3 {
             //  do a stat with a retry if not found
             if (s3_transport_ptr->is_last_file_to_close() && result.ok()) {
 
+                // Reset global variables.  These cached values for these variables are no longer
+                // valid once the last close is performed on the data object.
+                //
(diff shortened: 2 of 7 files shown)


packaging/test_irods_resource_plugin_s3.py
```diff
@@ -150,6 +150,9 @@ def test_put_get_small_file_in_repl_node(self):
     def test_put_get_large_file_in_repl_node(self):
         pass
 
+    @unittest.skipIf(True, 'test does not work in decoupled because we are using same bucket for multiple resources')
```
trel, line 153: `decoupled`?  or `detached`?

korydraughn, line 153: I think you can refactor this to just use `.skip("your message here ...")`.

Same for the others.

s3/s3_operations.cpp
```diff
@@ -998,6 +998,14 @@ namespace irods_s3 {
             //  do a stat with a retry if not found
             if (s3_transport_ptr->is_last_file_to_close() && result.ok()) {
 
+                // reset global variables
```
korydraughn, line 1001: Consider expanding on this.

It seems important for future readers to know why these need to be reset.

packaging/test_irods_resource_plugin_s3.py
JustinKyleJames: In this case decoupled is correct.  These are tests where we use decoupled naming so the key is the reverse object id.  That breaks tests that use the same bucket for multiple resources.

JustinKyleJames: I made the changes to the newly inserted lines.  I left the unchanged lines alone.

s3/s3_operations.cpp
JustinKyleJames: done

JustinKyleJames:
@korydraughn @trel  Is this ready to go?

korydraughn (approved):
Looks good to me.

Let's give @trel a chance to respond before squashing and adding the pound.

alanking (approved):
Looks good. Let's wait for @trel, like @korydraughn said.

trel:
yes looks good.   please squash and pound.

JustinKyleJames:
> yes looks good. please squash and pound.

done

PR #2127, opened 2023-07-25
[2122] Reset global vars after last file closes.

JustinKyleJames:
This is the same as the pull request opened for 4-2-stable.  It has gone through testing on main.

Update S3 repl test to use checksum.

Do not run S3 repl test when in decoupled mode as we are using the same bucket for all resources thus the object keys conflict.


korydraughn (approved):
Looks good to me.

Add the pound when ready.

alanking (approved):

JustinKyleJames:
> Looks good to me.
> 
> Add the pound when ready.

Done

Issue #2115, opened 2023-05-11
Move some python tests with very large files so they don't run in every suite.


JustinKyleJames:
- [x] main
- [x] 4-2-stable
---

There are a few S3 tests that deal with very large files.  Move these to a class that derives from Test_S3_NoCache_Base and only derive a limited number of suites from this new derived class.

These tests don't need to run in every single scenario (V4, V2, with SSE, etc.).  It takes too long and is redundant.

alanking:
@JustinKyleJames - Please close if complete. Thanks!

JustinKyleJames:
closing

Issue #2110, opened 2023-05-03
Upgrade pip to 20.3.4 before installing minio in test hook


JustinKyleJames:
- [x] main
- [x] 4-2-stable
---
Upgrade pip to 20.3.4 before installing minio in test hook.  This is required to install minio.

JustinKyleJames:
@alanking Do you know what needs to be done for this in main?  Is it just the same changes but with python2 replaced with python3?

Is the minio 6.0.2 version selection still required?

If you don't know I can play around with it and test it out.

alanking:
If things are working as expected in main, then we probably don't need to do anything. I'm guessing we needed to do this in 4-2-stable due to having to support older platforms, but the newer platforms in 4.3.x will have a new enough pip so this isn't an issue. Thoughts?

If agreed, please remove checkboxes (or just remove the "main" checkbox) and close. Thanks!

alanking:
Oops, I guess I could answer the question...

I think that if things aren't working in main, then we just need to upgrade to the latest minio that can be used on all the platforms. In short, no, I don't know what needs to be done for main. :)

alanking:
After some discussion, we decided that we should set 20.3.4 as the *minimum* version and @JustinKyleJames will update #2130 to reflect that.

JustinKyleJames:
I changed it to pip>=20.3.4.  This passed tests.

alanking:
@JustinKyleJames - Please checkbox and close. Thanks!

PR #2130, opened 2023-08-04
Issues 2110 2112 main

JustinKyleJames:
This is to address several issues:

- 2110 and 2112 are cherry-picked changes from 4-2-stable into main (with python2 replaced by python3).
- 2128 is needed for a recent change in iRODS core changing size_t to rodsLong_t in file_object::size.
- 2129 addresses something missed in the minio test suite when #2115 was implemented causing minio tests to fail.

I ran this through the testing environment and it passed.

### DIFF
--- a/irods_consortium_continuous_integration_test_hook.py
+++ b/irods_consortium_continuous_integration_test_hook.py
@@ -32,6 +32,7 @@ def get_package_type():
     return pt
 
 def install_test_prerequisites():
+    irods_python_ci_utilities.subprocess_get_output(['sudo', 'python3', '-m', 'pip', 'install', '--upgrade', 'pip>=20.3.4'], check_rc=True)
     irods_python_ci_utilities.subprocess_get_output(['sudo', 'python3', '-m', 'pip', 'install', 'boto3', '--upgrade'], check_rc=True)
     irods_python_ci_utilities.subprocess_get_output(['sudo', 'python3', '-m', 'pip', 'install', 'minio', '--upgrade'], check_rc=True)
     irods_python_ci_utilities.subprocess_get_output(['sudo', '-EH', 'python3', '-m', 'pip', 'install', 'unittest-xml-reporting==1.14.0'])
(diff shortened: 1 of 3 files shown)


irods_consortium_continuous_integration_test_hook.py
```diff
@@ -32,6 +32,7 @@ def get_package_type():
     return pt
 
 def install_test_prerequisites():
+    irods_python_ci_utilities.subprocess_get_output(['sudo', 'python3', '-m', 'pip', 'install', '--upgrade', 'pip==20.3.4'], check_rc=True)
```
korydraughn, line 35: Is 20.3.4 a hard requirement?

What happens if we didn't specify a specific version?
Does it upgrade to a version that breaks testing?

JustinKyleJames: This was a change made by Alan.  I am not sure of the reasoning.

korydraughn: @alanking Thoughts?

alanking: I believe I determined this to be the earliest `pip` version that I could use to install the appropriate `minio` client. We may not need to pin the version (or at least make it something more recent) for 4.3+.

alanking: Given that the change exists in 4-2-stable and not main, I'm fairly certain we don't need this. However, I'm not opposed to upgrading `pip` as part of this process because it sometimes causes problems if we don't.

alanking (commented):
Seems good to me

irods_consortium_continuous_integration_test_hook.py
korydraughn: Pinning it doesn't sound bad. Doing that will allow us to know when the minimum version isn't sufficient. That also means we'll always be forced to know which version of pip works for the software.

JustinKyleJames: In reference to Alan's comment below, I cherry-picked and tested this on 4.3 because issue #2110 has main in the task list.

JustinKyleJames: I changed the test hook to use pip>=20.3.4.  This passed tests in the testing environment.

Note I went ahead and squashed that one commit with the original since it was so minor.

alanking (approved):
Looks good to me. Add #s after one more approval

korydraughn (approved):
Add the pounds

PR #2132, opened 2023-08-14
[2124] Set the proper open flags on read for checksum after put. (4-2-stable)

JustinKyleJames:
Added test cases for large file checksum with MPU disabled. Placed test case in a newly created class so that it only runs for MPU disabled.

### DIFF
--- a/packaging/resource_suite_s3_nocache.py
+++ b/packaging/resource_suite_s3_nocache.py
@@ -2759,3 +2759,28 @@ class Test_S3_NoCache_Glacier_Base(session.make_sessions_mixin([('otherrods', 'r
             s3plugin_lib.remove_if_exists(file2)
             s3plugin_lib.remove_if_exists(file1_get)
             s3plugin_lib.remove_if_exists(file2_get)
+
+class Test_S3_NoCache_MPU_Disabled_Base(Test_S3_NoCache_Base):
+
+    def test_iput_large_file_with_checksum_issue_2124(self):
+
+        try:
+
+            file1 = "issue_2124_file"
+            file1_get = "issue_2124_file.get"
+            file1_size = 32*1024*1024 + 1
+
+            # create and put file
+            lib.make_arbitrary_file(file1, file1_size)
+
+            self.user0.assert_icommand("iput -K " + file1)  # iput
+            with open(file1, 'rb') as f:
+                checksum = base64.b64encode(hashlib.sha256(f.read()).digest()).decode()
+            self.user0.assert_icommand("ils -L", 'STDOUT_SINGLELINE', "sha2:" + checksum)  # check proper checksum
+
+        finally:
+
+            # cleanup
+            self.user0.assert_icommand("irm -f {file1}".format(**locals()), 'EMPTY')
+            s3plugin_lib.remove_if_exists(file1)
+            s3plugin_lib.remove_if_exists(file1_get)
--- a/s3/s3_operations.cpp
+++ b/s3/s3_operations.cpp
@@ -78,6 +78,7 @@ namespace irods_s3 {
         std::ios_base::openmode open_mode;
         std::shared_ptr<dstream> dstream_ptr;
         std::shared_ptr<s3_transport> s3_transport_ptr;
+        bool checked_for_checksum_read_after_put;
     }; // end per_thread_data
 
     class fd_to_data_map {
@@ -751,9 +752,12 @@ namespace irods_s3 {
             rodsLog(developer_messages_log_level, "%s:%d (%s) [[%lu]] oprType set to %d\n",
                     __FILE__, __LINE__, __FUNCTION__, thread_id, oprType);
 
-            // fix open mode
             ios_base::openmode open_mode;
             if (oprType == PUT_OPR) {
+                // Fix open mode so we can stream
+                // The open mode is forced to O_CREAT, O_WRONLY, O_TRUNC
+                // If this follows with a read, then we are doing a read after write for checksum, the open_mode
+                // will be set back to the original value.
                 open_mode = translate_open_mode_posix_to_stream(O_CREAT | O_WRONLY | O_TRUNC, __FUNCTION__);
             } else {
                 open_mode = translate_open_mode_posix_to_stream(file_obj->flags(), __FUNCTION__);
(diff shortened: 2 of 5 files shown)


packaging/resource_suite_s3_nocache.py
```diff
@@ -2759,3 +2759,28 @@ def test_put_get_glacier_ir(self):
...
+
+class Test_S3_NoCache_MPU_Disabled_Base(Test_S3_NoCache_Base):
+
+    def test_iput_large_file_with_checksum_issue_2124(self):
+
+        try:
+
+            file1 = "issue_2124_file"
+            file1_get = "issue_2124_file.get"
+            file1_size = 32*1024*1024 + 1
```
alanking, line 2771: Maybe add a comment or even some words to the test name itself to explain why it needs to be 1 byte above 32MB?

s3/s3_operations.cpp
```diff
@@ -845,6 +849,55 @@ namespace irods_s3 {
...
+            // would work.  We need to check if this was a PUT_OPR and if so we have moved on to the read-only portion of
+            // processing for checksum calculation.  If that is the case, the open_mode will be reset to O_RDONLY.
+            // So that this isn't done on every read call, a flag (checked_for_checksum_read_after_put) is set to indicate that we
+            // have already checked for this condition and adjusted the open_mode accordingly.
+            if (!data.checked_for_checksum_read_after_put) {
+
+                // Get oprType
+                int oprType = -1;
+                bool found = false;
+                for (int i = 0; i < NUM_L1_DESC; ++i) {
```
alanking, line 871: Consider using a `std::span` here as a "view" of the L1 descriptor table. This would also allow you to use a range-based for loop.

Feel free to ignore this as what you're doing here is standard practice (for now)

s3/s3_operations.cpp
```diff
@@ -845,6 +849,55 @@ namespace irods_s3 {
...
+                   if (L1desc[i].inuseFlag) {
+                       if (L1desc[i].dataObjInp && L1desc[i].dataObjInfo &&
+                               L1desc[i].dataObjInp->objPath == file_obj->logical_path()
+                               && L1desc[i].dataObjInfo->filePath == file_obj->physical_path()) {
```
alanking, line 875: Does the `dataObjInfo` have the `objPath` here? Could avoid the check for `dataObjInp` if so.

s3/s3_operations.cpp
```diff
@@ -845,6 +849,55 @@ namespace irods_s3 {
...
+
+                           found = true;
+                           oprType = L1desc[i].dataObjInp->oprType;
```
alanking, line 878: Could we just `break` here instead of keeping track of `found` and `break`ing on the next iteration?

s3/s3_operations.cpp
```diff
@@ -845,6 +849,55 @@ namespace irods_s3 {
...
+                       }
+                   } else if (found) {
+                       break;
+                   }
+                }
+                
+                data.checked_for_checksum_read_after_put = true;
+
+                // If oprType is PUT_OPR, adjust the open mode to O_RDONLY.
+                if (oprType == PUT_OPR) {
```
alanking, line 888: Is it a problem if `oprType == -1` at this point? That might mean that the L1 descriptor is invalid, right?

s3/s3_operations.cpp
```diff
@@ -845,6 +849,55 @@ namespace irods_s3 {
...
+
+                    std::ios_base::openmode open_mode;
+                    open_mode = translate_open_mode_posix_to_stream(O_RDONLY, __FUNCTION__);
+
+                    // set open_mode in per_thread_data
+                    data.open_mode = open_mode;
```
korydraughn, line 894: These lines can be collapsed into a single line.
```c++
data.open_mode = translate_open_mode_posix_to_stream(O_RDONLY, __FUNCTION__);
```

s3/s3_operations.cpp
```diff
@@ -78,6 +78,7 @@ namespace irods_s3 {
         std::ios_base::openmode open_mode;
         std::shared_ptr<dstream> dstream_ptr;
         std::shared_ptr<s3_transport> s3_transport_ptr;
+        bool checked_for_checksum_read_after_put;
```
korydraughn, line 81: Consider initializing the boolean.

s3/s3_operations.cpp
korydraughn: Sadly, we can't use `std::span` here. It's a C++20 feature.

s3/s3_operations.cpp
```diff
@@ -751,9 +752,12 @@ namespace irods_s3 {
             rodsLog(developer_messages_log_level, "%s:%d (%s) [[%lu]] oprType set to %d\n",
                     __FILE__, __LINE__, __FUNCTION__, thread_id, oprType);
 
-            // fix open mode
             ios_base::openmode open_mode;
             if (oprType == PUT_OPR) {
+                // Fix open mode so we can stream
+                // The open mode is forced to O_CREAT, O_WRONLY, O_TRUNC
+                // If this follows with a read, then we are doing a read after write for checksum, the open_mode
+                // will be set back to the original value.
```
korydraughn, line 760: I'm a little confused by the last two lines of the comment. Can you provide more explanation here in the comments?

s3/s3_operations.cpp
alanking: D'oh. 4-2-stable

packaging/resource_suite_s3_nocache.py
JustinKyleJames: Well, partially it is because I don't remember if the cutoff for parallel transfer is >= 32 MB or > 32MB.

But I will put a comment in there.

s3/s3_operations.cpp
JustinKyleJames: I think it would make sense to continue since this check is a fix for a specific scenario meaning if we encounter anything else we just proceed normally.

That may be an indication of a problem but it isn't related to what we are doing here.

s3/s3_operations.cpp
JustinKyleJames: done

s3/s3_operations.cpp
JustinKyleJames: Are you asking for me to add a constructor / initialization list for this struct?

korydraughn: No, you don't have to add a constructor. You can initialize the bool directly.

You may want to initialize the open_mode too.

Keep in mind I don't how the struct is handled in the plugin. If you can guarantee that the struct is never in an invalid state, then maybe none of this is necessary.

trel:
new one at https://github.com/irods/irods_resource_plugin_s3/pull/2133

PR #2133, opened 2023-08-16
[2124] Be selective when setting open flags on PUT_OPR in s3_file_open_operation() (4-2-stable)

JustinKyleJames:
Added test cases for large file checksum with MPU disabled. Placed test case in a newly created class so that it only runs for MPU disabled.

Note that this replaces pull request #2132 as I went with a different/simpler approach.

All tests passed.

### DIFF
--- a/s3/s3_operations.cpp
+++ b/s3/s3_operations.cpp
@@ -751,10 +751,23 @@ namespace irods_s3 {
             rodsLog(developer_messages_log_level, "%s:%d (%s) [[%lu]] oprType set to %d\n",
                     __FILE__, __LINE__, __FUNCTION__, thread_id, oprType);
 
-            // fix open mode
             ios_base::openmode open_mode;
-            if (oprType == PUT_OPR) {
-                open_mode = translate_open_mode_posix_to_stream(O_CREAT | O_WRONLY | O_TRUNC, __FUNCTION__);
+
+            // Update open_mode when oprType=PUT_OPR.  There are three scenarios to consider: 
+            //
+            //   1.  The mode is set to O_WRONLY.  This would not stream because the O_CREAT or
+            //       O_TRUNC flag are not set.  Update the open flag to O_WRONLY | O_CREAT | O_TRUNC 
+            //       as we know a PUT_OPR is always a full file write or overwrite.
+            //
+            //   2.  The mode is set to O_RDWR.  This happens when there is a write
+            //       which will be followed up by a read for the checksum.  This would not
+            //       allow streaming because the file is opened in read and write mode. 
+            //       As before, update the open flag to O_WRONLY | O_CREAT | O_TRUNC.
+            //
+            //   2.  The mode is set to O_RDONLY.  This is the read for checksum that follows the
+            //       write.   Leave the oprType alone in this scenario. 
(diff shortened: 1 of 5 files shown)


s3/s3_operations.cpp
```diff
@@ -751,10 +751,23 @@ namespace irods_s3 {
...
+            // Update open_mode when oprType=PUT_OPR.  There are three scenarios to consider: 
+            //
+            //   1.  The mode is set to O_WRONLY.  This would not stream because the O_CREAT or
+            //       O_TRUNC flag are not set.  Update the open flag to O_WRONLY | O_CREAT | O_TRUNC 
+            //       as we know a PUT_OPR is always a full file write or overwrite.
+            //
+            //   2.  The mode is set to O_RDWR.  This happens when there is a write
+            //       which will be followed up by a read for the checksum.  This would not
+            //       allow streaming because the file is opened in read and write mode. 
+            //       As before, update the open flag to O_WRONLY | O_CREAT | O_TRUNC.
```
alanking, line 765: To be clear, we won't lose the fact that `O_RDWR` is set because the open flags on `file_obj` are separate from the open flags being used with `open_mode`? In other words, there's no need to "restore" the `O_RDWR` because it's not being overwritten here, right?

alanking (commented):
Very nice. Comments look good, code looks good. Just had one question to make sure I'm understanding correctly.

s3/s3_operations.cpp
JustinKyleJames: To my knowledge the O_RDWR isn't necessary in this situation.  When the read for checksum happens the flag is set to O_RDONLY.

It might make sense in a POSIX environment but as far as I can tell the O_RDWR isn't needed at this point.

If someone opens the file with a real O_RDWR I do not believe the oprType would be set to PUT_OPR.

JustinKyleJames: To add some context, multiple threads open the file.  The first is O_RDWR.  The others are O_WRONLY.  This is all fine for POSIX but the S3 plugin doesn't share this information between threads and the thread that opens with O_RDWR thinks that it can't stream because it appears to that thread that the thread is going to be doing some random access reads and writes.

s3/s3_operations.cpp
```diff
@@ -751,10 +751,23 @@ namespace irods_s3 {
...
+            //
+            //   2.  The mode is set to O_RDONLY.  This is the read for checksum that follows the
+            //       write.   Leave the oprType alone in this scenario. 
+            if (oprType == PUT_OPR && (file_obj->flags() & O_WRONLY || file_obj->flags() & O_RDWR)) {
```
korydraughn, line 769: Bitwise operations are a little weird when dealing with logical operators.

Please verify the logic presented behaves as you expect. I think issues with precedence could be at play here.

You can do that using compiler explorer.

JustinKyleJames: > Bitwise operations are a little weird when dealing with logical operators.
> 
> Please verify the logic presented behaves as you expect. I think issues with precedence could be at play here.
> 
> You can do that using compiler explorer.

I've tested it out and it is all good.  Also verified the operator precedence.  Bitwise and evaluates before logical or.

korydraughn: Excellent.

korydraughn (approved):
No obvious issues to me.

Please add the pounds when ready.

Issue #2147, opened 2023-11-25
Do not initialize `std::ios_base::openmode` objects with an int


SwooshyCueb:
In a couple of different places, we attempt to instantiate `std::ios_base::openmode` objects with an `int`:
https://github.com/irods/irods_resource_plugin_s3/blob/d4904e42bb2ac07e165d36b649472a3a8b991f4d/s3/s3_transport/include/s3_transport.hpp#L231
https://github.com/irods/irods_resource_plugin_s3/blob/d4904e42bb2ac07e165d36b649472a3a8b991f4d/s3/s3_operations.cpp#L434
The underlying type of `std::ios_base::openmode` is *implementation defined*. This causes a number of errors when building against libstdc++.

alanking:
Probably no use doing this for 4-2-stable as it is not going to impact behavior, right?

korydraughn:
Yeah, no need to change 4-2-stable for this.

PR #2148, opened 2023-11-25
[#2147] use static_cast when instantiating ios_base::openmode with an int

SwooshyCueb:
Addresses #2147

Still needs to be tested against libc++

### DIFF
--- a/s3/s3_operations.cpp
+++ b/s3/s3_operations.cpp
@@ -431,7 +431,7 @@ namespace irods_s3 {
                 (oflag & O_ACCMODE) == O_WRONLY, (oflag & O_ACCMODE) == O_RDWR, (oflag & O_ACCMODE) == O_RDONLY,
                 (oflag & O_TRUNC) != 0, (oflag & O_CREAT) != 0, (oflag & O_APPEND) != 0);
 
-        ios_base::openmode mode = 0;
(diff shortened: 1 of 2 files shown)


s3/s3_operations.cpp
```diff
@@ -431,7 +431,7 @@ namespace irods_s3 {
                 (oflag & O_ACCMODE) == O_WRONLY, (oflag & O_ACCMODE) == O_RDWR, (oflag & O_ACCMODE) == O_RDONLY,
                 (oflag & O_TRUNC) != 0, (oflag & O_CREAT) != 0, (oflag & O_APPEND) != 0);
 
-        ios_base::openmode mode = 0;
+        ios_base::openmode mode = static_cast<ios_base::openmode>(0);
```
korydraughn, line 434: This is fine, but I'm just now noticing we don't guard against invalid values for `oflag` (i.e. invalid access mode).

This can also be handled in a separate PR.

SwooshyCueb:
Works fine building against libc++ too, marking as ready for review

s3/s3_operations.cpp
SwooshyCueb: Should I open an issue for it?

korydraughn (approved):
I think this is safe.

Pound it.

SwooshyCueb:
\#'d

PR #2152, opened 2023-11-30
[#2138] Absorb libs3 as an in-repo object library

SwooshyCueb:
Addresses #2138

Once #2149 is merged, this PR will be rebased and the first commit will disappear.

For some reason, clang will sometimes crash when compiling the newly-added libs3 code. I can't figure out why or how to reliably reproduce it.

PR #2155, opened 2023-12-14
Issues 2151 2153

JustinKyleJames:
This addresses the istream append issues.

I put this on top of #2152.  Once that is merged this will be rebased.

All tests passed.

### DIFF
--- a/packaging/resource_suite_s3_nocache.py
+++ b/packaging/resource_suite_s3_nocache.py
@@ -1910,6 +1910,41 @@ OUTPUT ruleExecOut
                 'DATA_REPL_STATUS', '0'])
             self.user0.run_icommand(['irm', '-f', logical_path])
 
+    def test_istream_append__issue_2153(self):
+        replica_number_in_s3 = 0
+        filename = 'test_istream_append__issue_2153.txt'
+        logical_path = os.path.join(self.user0.session_collection, filename)
+        content = 'this is the original file for istream'
+        content_append1 = '\nline1 appended to the file'
+        content_append2 = '\nline2 appended to the file'
+
+        try:
+            # istream to a new data object and ensure that it is created successfully.
+            self.user0.assert_icommand(['istream', 'write', logical_path], input=content)
+            self.assertEqual(str(1), lib.get_replica_status(self.user0, filename, replica_number_in_s3))
+
+            # append to the file
+            self.user0.assert_icommand(['istream', 'write', '-a', logical_path], input=f'{content_append1}')
+
+            # append to the file again
+            self.user0.assert_icommand(['istream', 'write', '-a', logical_path], input=f'{content_append2}')
+
+            # verify the file size and repl status in ils -l 
+            self.user0.assert_icommand(['ils', '-l', logical_path], 'STDOUT', str(len(content) + len(content_append1) + len(content_append2))) 
+            self.assertEqual(str(1), lib.get_replica_status(self.user0, filename, replica_number_in_s3))
+
+            # Ensure that the replica actually contains the contents streamed into it.
+            self.user0.assert_icommand(['istream', 'read', logical_path], 'STDOUT', [content, content_append1, content_append2])
+
+        finally:
+            # Set the replica status here so that we can remove the object even if it is stuck in the locked status.
+            print(self.user0.run_icommand(['ils', '-l'])[0])   # just debug
+            self.admin.run_icommand([
+                'iadmin', 'modrepl',
+                'logical_path', logical_path,
+                'replica_number', str(replica_number_in_s3),
+                'DATA_REPL_STATUS', '0'])
+            self.user0.run_icommand(['irm', '-f', logical_path])
 
     def test_iput_with_invalid_secret_key_and_overwrite__issue_6154(self):
         replica_number_in_s3 = 0
--- a/s3/s3_operations.cpp
+++ b/s3/s3_operations.cpp
@@ -441,7 +441,7 @@ namespace irods_s3 {
             mode |= ios_base::in;
         }
 
-        if ((oflag & O_TRUNC) || (oflag & O_CREAT)) {
+        if ((oflag & O_TRUNC)) {
             mode |= ios_base::trunc;
         }
 
@@ -449,6 +449,19 @@ namespace irods_s3 {
             mode |= ios_base::app;
             mode &= ~ios_base::trunc;  // turn off trunc flag
         }
+
+        logger::debug("{}:{} ({}) [[{}]] translated open mode is [app={}][binary={}][in={}][out={}][trunc={}][ate={}]",
+                __FILE__,
+                __LINE__,
+                __FUNCTION__,
+                thread_id, 
+                (mode & std::ios::app) != 0,
+                (mode & std::ios::binary) != 0,
+                (mode & std::ios::in) != 0,
+                (mode & std::ios::out) != 0,
+                (mode & std::ios::trunc) != 0,
+                (mode & std::ios::ate) != 0
+                );
 
         return mode;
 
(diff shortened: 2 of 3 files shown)


alanking:
@JustinKyleJames - #2152 is now merged. Let's go ahead and rebase.

s3/s3_transport/include/s3_transport.hpp
```diff
@@ -836,6 +835,29 @@ namespace irods::experimental::io::s3_transport
             return existing_object_size_;
         }
 
+        bool is_open_cache_file() {
+            if (use_cache_) {
```
alanking, line 839: Same thing here - this could be flattened by reversing the logic:
```c++
if (!use_cache_) {
    return false;
}
// ...the rest of the stuff
```

Also, I think the name could use a tweak, unless I'm misunderstanding. Maybe one of these...
- is_cache_file_open
- has_open_cache_file
- cache_file_is_open

Just a semantics thing - the stream itself is not an "open cache file". Not a big deal.

alanking (commented):
Looks good overall. Had a couple of pretty minor suggestions.

s3/s3_transport/include/s3_transport.hpp
JustinKyleJames: Flipped the logic and renamed the method to `is_cache_file_open()`.

s3/s3_operations.cpp
```diff
@@ -1334,58 +1334,57 @@ namespace irods_s3 {
...
                 }
-            }
 
-            // there is not an open cache file, do the normal HEAD to S3 
-            return s3_file_stat_operation_with_flag_for_retry_on_not_found(_ctx, _statbuf, false);
-        } else {
-            return s3_file_stat_operation_with_flag_for_retry_on_not_found(_ctx, _statbuf, false);
+                auto result = SUCCESS();
+                result.code(status);
+                return result;
```
alanking, line 1382: Feels a little dirty to suggest this, but `SUCCESS()` is a shortcut for `CODE(0)`. So...

```suggestion
                return CODE(status);
```

https://github.com/irods/irods/blob/aff93a3fbe417c8ec2e9d834b5eb2d6b2ae31083/lib/core/include/irods/irods_error.hpp#L133-L134

Please ignore if you feel like it. :p

alanking (commented):
Saw one more thing - take it or leave it. This seems good to me. Maybe wait for one more reviewer before we put it in.

s3/s3_operations.cpp
JustinKyleJames: I went ahead and made the change.

korydraughn (commented):
Looks good overall.

packaging/resource_suite_s3_nocache.py
```diff
@@ -1910,6 +1910,39 @@ def test_istream_nonexistent_file__issue_6479(self):
...
+            # Ensure that the replica actually contains the contents streamed into it.
+            self.user0.assert_icommand(['iget', logical_path, '-'], 'STDOUT_MULTILINE', [content, content_append1, content_append2])
+
+        finally:
+            # Set the replica status here so that we can remove the object even if it is stuck in the locked status.
+            self.admin.run_icommand([
+                'iadmin', 'modrepl',
+                'logical_path', logical_path,
+                'replica_number', str(replica_number_in_s3),
+                'DATA_REPL_STATUS', '0'])
```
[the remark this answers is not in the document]
alanking: I feel strongly that we should clean up the data object, and this is the way to do it. The test is reading and writing the data object so any problems will `raise` an `AssertionError`, so changing the replica status in order to clean up will not mask the fact that something went wrong above.

At best, not removing the object will allow the person running the test to inspect the object later, which wouldn't work in an ephemeral environment (e.g. k8s, no `--leak-containers` in testing environment, etc.) and will cause other tests to fail that may have otherwise passed. 

If we want to see more information on failure, the same thing can be accomplished by `print`ing an `ils -L` on the logical path or parent collection right before we do this as a debugging effort. That would work better in an ephemeral environment (i.e. shows up in the logs) and allows other tests to run.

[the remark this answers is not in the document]
korydraughn: I agree with all of that, but if things are working as intended, `iadmin modrepl` isn't needed.

I can see this going both ways. Let's find a few minutes to discuss today.

packaging/resource_suite_s3_nocache.py
```diff
@@ -1910,6 +1910,39 @@ def test_istream_nonexistent_file__issue_6479(self):
...
+        try:
+            # istream to a new data object and ensure that it is created successfully.
+            self.user0.assert_icommand(['istream', 'write', logical_path], input=content)
+            self.assertEqual(str(1), lib.get_replica_status(self.user0, filename, replica_number_in_s3))
+
+            # append to the file
+            self.user0.assert_icommand(['istream', 'write', '-a', logical_path], input=f'\n{content_append1}')
+
+            # append to the file again
+            self.user0.assert_icommand(['istream', 'write', '-a', logical_path], input=f'\n{content_append2}')
```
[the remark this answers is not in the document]
JustinKyleJames: The reasons I did it that way is because of the assertion in 1936.  If I put the newline in those it searches for lines that have the newline character in it and fails.

If there is a better way to approach this then I would be happy to make a change.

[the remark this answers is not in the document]
JustinKyleJames: Kory pointed out that using 'STDOUT' vs 'STDOUT_MULTILINE' handles the newline issue so I made the change so the vars now have the newline in it.

packaging/resource_suite_s3_nocache.py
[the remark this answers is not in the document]
JustinKyleJames: I made the change to check the repl status so there is no chance for this failure to be undetected.  I kept the modrepl so in case there was a failure it doesn't cascade to create a lot of false positives.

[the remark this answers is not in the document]
korydraughn: Sounds good.

korydraughn:
Will review once the latest changes appear.

packaging/resource_suite_s3_nocache.py
```diff
@@ -1915,25 +1915,26 @@ def test_istream_append__issue_2153(self):
...
+            self.user0.assert_icommand(['istream', 'write', '-a', logical_path], input=f'{content_append2}')
 
             # verify the file size in ils -l - added two newlines above thus the +2 below
-            self.user0.assert_icommand(['ils', '-l', logical_path], 'STDOUT', str(len(content) + len(content_append1) + len(content_append2)+ 2)) 
+            self.user0.assert_icommand(['ils', '-l', logical_path], 'STDOUT', str(len(content) + len(content_append1) + len(content_append2))) 
 
             # Ensure that the replica actually contains the contents streamed into it.
-            self.user0.assert_icommand(['iget', logical_path, '-'], 'STDOUT_MULTILINE', [content, content_append1, content_append2])
+            self.user0.assert_icommand(['istream', 'read', logical_path], 'STDOUT', [content, content_append1, content_append2])
+            #self.user0.assert_icommand(['iget', logical_path, '-'], 'STDOUT_MULTILINE', [content, content_append1, content_append2])
```
alanking, line 1937: Torch?

packaging/resource_suite_s3_nocache.py
```diff
@@ -1915,25 +1915,26 @@ def test_istream_append__issue_2153(self):
...
 
         finally:
```
alanking, line 1939: I think we agreed to put an `ils -L` debug or something at the start of the finally block?

alanking (commented):
Couple of lingering things then we can get this in

packaging/resource_suite_s3_nocache.py
JustinKyleJames: Forgot to push after the latest change.

alanking (approved):
Nice. Please squash to taste and add #s.

JustinKyleJames:
> Nice. Please squash to taste and add #s.

Done

Issue #2146, opened 2023-11-08
Decoupled mode is not being honored in cacheless mode after redirect


JustinKyleJames:
- [x] main
- [x] 4-3-stable
---

When a client is connected to one server and the S3 resource is attached to another server and in decoupled mode, an iput does not honor decoupled mode.  It instead uses the path in iRODS for the key.

The object can still be read from any of the servers since the key that is used in S3 is stored in the catalog.

I tested the following scenarios:

* Client connected to provider, resource attached to provider, iput wrote the file correctly with the numeric decoupled key.
* Client connected to consumer, resource attached to provider, iput wrote the file incorrectly using the iRODS path as the key.
* Client connected to provider, resource attached to consumer, iput wrote the file incorrectly using the iRODS path as the key.
* Client connected to consumer, resource attached to consumer,  iput wrote the file correctly with the numeric decoupled key.

JustinKyleJames:
It appears that the problem is caused by the L1desc entry not following a redirect.  The data id is used for decoupled mode and it is taken from the L1desc table.

JustinKyleJames:
Note in the archive mode, the archive resource must be attached to the same server as the cache resource.  So it is possible that that is why it works because once you have the cache file and you are syncing to archive, you are not doing any more redirects.

JustinKyleJames:
This is not an issue with archive mode as the data object id is read directly from the irods::file_object class since the object has already been written to cache and therefore the database.  This is not yet available in cacheless mode so we had to use the L1desc table.

JustinKyleJames:
> Hi,
> 
> Is there any update regarding this issue? Is there any plan to pick it up in a coming release?
> 
> TIA for your prompt response.

The fix will be in the next release of the plugin.  We should be able to fix it in the next few weeks.

korydraughn:
The fix would apply to 4.2 and 4.3 assuming it doesn't require any changes to the iRODS server.

JustinKyleJames:
The commits have been merged to both main and 4-2-stable.  Closing this issue.

JustinKyleJames:
Reopening this issue because the fix did not work on 4.2.12.  When I tested it I accidentally had the resource set to detached mode so the redirect did not happen as intended.

[...]

PR #2159, opened 2024-01-06
Issue 2146 main

JustinKyleJames:
This fixes issue #2146. All tests passed.

 In addition, I added two general test cases to test switching archive naming policy consistent->decoupled and decoupled->consistent after object had already been created. These were added for the following reasons:

1. This is something we should be testing since the physical path in the DB for existing objects should override the archive naming policy.
2. On my first attempt to fix 2146 I broke this behavior and realized there was no specific test for this.

See 03eb431b059802458a2d2ffde02ecdc636295e50 which fixes the scenario I was trying to fix with the open flags.  I am now checking the openType.  This is now working as expected.

### DIFF
--- a/packaging/resource_suite_s3_nocache.py
+++ b/packaging/resource_suite_s3_nocache.py
@@ -119,8 +119,6 @@ class Test_S3_NoCache_Base(session.make_sessions_mixin([('otherrods', 'rods')], 
         if hasattr(self, 's3sse'):
             self.s3_context += ';S3_SERVER_ENCRYPT=' + str(self.s3sse)
 
-        self.s3_context += ';ARCHIVE_NAMING_POLICY=' + self.archive_naming_policy
-
         self.admin.assert_icommand("iadmin modresc demoResc name origResc", 'STDOUT_SINGLELINE', 'rename', input='yes\n')
 
         self.admin.assert_icommand("iadmin mkresc demoResc s3 " + hostname + ":/" + self.s3bucketname + "/demoResc " + self.s3_context, 'STDOUT_SINGLELINE', 's3')
@@ -2869,3 +2867,39 @@ class Test_S3_NoCache_MPU_Disabled_Base(Test_S3_NoCache_Base):
             self.user0.assert_icommand("irm -f {file1}".format(**locals()), 'EMPTY')
             s3plugin_lib.remove_if_exists(file1)
             s3plugin_lib.remove_if_exists(file1_get)
+
+class Test_S3_NoCache_Decoupled_Base(Test_S3_NoCache_Base):
+
+    def test_decoupled_redirect_issue_2146(self):
+
+        file1 = "f1"
+        file1_size = 2*1024*1024
+        retrieved_file = "f1.get"
+        resource_host = test.settings.HOSTNAME_3
+        resource_name = "s3_resc_on_host3"
+
+        # create an S3 resource
+        self.admin.assert_icommand(f'iadmin mkresc {resource_name} s3 {resource_host}:/{self.s3bucketname}/{resource_name} {self.s3_context}', 'STDOUT_SINGLELINE', 's3')
+
+        try:
+            # create the file
+            lib.make_arbitrary_file(file1, file1_size)
+
+            # put the file
+            self.user1.assert_icommand(f'iput -R {resource_name} {file1}')  # iput
+
+            # get and verify the file contents
+            self.user1.assert_icommand(f'iget {file1} {retrieved_file}')
+
+            self.assertTrue(filecmp.cmp(file1, retrieved_file))  # confirm retrieved is correct
+
+            # verify in the physical path that decoupled mode was used
+            print(self.user1.run_icommand(['ils', '-l', file1])[0])   # just debug
+            self.user1.assert_icommand(f'ils -L {file1}', 'STDOUT_SINGLELINE', f'/{self.s3bucketname}/[0-9]+/{file1}', use_regex=True)
+
+        finally:
+            # cleanup
+            self.user1.assert_icommand("irm -f %s" % file1)  # irm
+            self.admin.assert_icommand("iadmin rmresc %s" % resource_name)
+            s3plugin_lib.remove_if_exists(file1)
+            s3plugin_lib.remove_if_exists(retrieved_file)
--- a/packaging/resource_suite_s3_nocache.py
+++ b/packaging/resource_suite_s3_nocache.py
@@ -2903,3 +2903,121 @@ class Test_S3_NoCache_Decoupled_Base(Test_S3_NoCache_Base):
             self.admin.assert_icommand("iadmin rmresc %s" % resource_name)
             s3plugin_lib.remove_if_exists(file1)
             s3plugin_lib.remove_if_exists(retrieved_file)
+
+    # This verifies that once the file is written to the DB, whatever mode was used remains in effect as the DB always wins.
+    def test_resource_updated_from_consistent_to_decoupled_issue_2161(self):
+
+        file1 = "f1"
+        file1_size = 2*1024
+        file2 = "f2"
+        file2_size = 3*1024
+        retrieved_file = 'f.get'
+        hostname = lib.get_hostname()
+        resource_name = "s3_resc"
+
+        s3_context_consistent = self.s3_context.replace('ARCHIVE_NAMING_POLICY=decoupled', 'ARCHIVE_NAMING_POLICY=consistent')
+        s3_context_decoupled = self.s3_context
+
+        # create the S3 resource
+        self.admin.assert_icommand(f'iadmin mkresc {resource_name} s3 {hostname}:/{self.s3bucketname}/{resource_name} {s3_context_consistent}', 'STDOUT_SINGLELINE', 's3')
+
+        try:
+            # create file1 and file2
+            lib.make_arbitrary_file(file1, file1_size)
+            lib.make_arbitrary_file(file2, file2_size)
+
+            # put file1
+            self.user1.assert_icommand(f'iput -R {resource_name} {file1}')
+
+            # get and verify the file contents
+            self.user1.assert_icommand(f'iget {file1} {retrieved_file}')
+            self.assertTrue(filecmp.cmp(file1, retrieved_file))  # confirm retrieved is correct
+
+            # verify in the physical path that consistent mode was used
+            print(self.user1.run_icommand(['ils', '-l', file1])[0])   # just debug
+            self.user1.assert_icommand(f'ils -L {file1}', 'STDOUT_SINGLELINE', f'/{self.s3bucketname}/{resource_name}/home/.+/{file1}', use_regex=True)
+
+            # update the resource to decoupled mode
+            self.admin.run_icommand(f'iadmin modresc {resource_name} context {s3_context_decoupled}')
+
+            # get and verify the file contents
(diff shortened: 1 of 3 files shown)


packaging/resource_suite_s3_nocache.py
```diff
@@ -2869,3 +2867,149 @@ def test_iput_large_file_with_checksum_issue_2124(self):
             self.user0.assert_icommand("irm -f {file1}".format(**locals()), 'EMPTY')
             s3plugin_lib.remove_if_exists(file1)
             s3plugin_lib.remove_if_exists(file1_get)
+
+class Test_S3_NoCache_Decoupled_Base(Test_S3_NoCache_Base):
+
+    def test_decoupled_redirect_issue_2146(self):
```
korydraughn, line 2873: How is this triggering a redirect? Are there multiple iRODS servers in the zone?

korydraughn (commented):
Looks good.

s3/s3_operations.cpp
```diff
@@ -2231,6 +2266,16 @@ namespace irods_s3 {
...
     irods::error s3_notify_operation( irods::plugin_context& _ctx,
         const std::string* str ) {
+        if (is_cacheless_mode(_ctx.prop_map())) {
+            // Go ahead and update the physical_path before a potential
+            // redirect so the DB gets updated properly but only do so if
+            // we are creating a new object. Otherwise, just use the path
+            // that already existed as what is in the DB must be honored.
+            irods::file_object_ptr object = boost::dynamic_pointer_cast<irods::file_object>(_ctx.fco());
+            if (object->flags() & O_CREAT) {
+                update_physical_path_for_decoupled_naming(_ctx);
```
alanking, line 2276: Does the O_CREAT flag get removed if it is used as part of, e.g. an overwrite via `iput -f`? Or is that API dependent? In other words, Is the check here sufficient to determine whether a new object is being created? Or maybe `update_physical_path_for_decoupled_naming` figures it out?

packaging/resource_suite_s3_nocache.py
alanking: I'm not sure whether plugin tests can be made topology-aware (perhaps when running manually), but the `resource_host` is set to `test.settings.HOSTNAME_3` which would guarantee that the client would not be connected to that host in the tests. But like you say, that only applies if `test.settings.ICAT_HOSTNAME`, `test.settings.HOSTNAME_1`, and `test.settings.HOSTNAME_3` are referring to different servers.

And there is of course no harm in testing it on a single machine as well. :)

packaging/resource_suite_s3_nocache.py
```diff
@@ -2869,3 +2867,149 @@ def test_iput_large_file_with_checksum_issue_2124(self):
...
+
+            # overwrite file1 with file2
+            self.user1.assert_icommand(f'iput -f -R {resource_name} {file2} {file1}')
+
+            # verify the path remains in consistent mode (DB rules) 
+            print(self.user1.run_icommand(['ils', '-l', file1])[0])   # just debug
+            self.user1.assert_icommand(f'ils -L {file1}', 'STDOUT_SINGLELINE', f'/{self.s3bucketname}/{resource_name}/home/.+/{file1}', use_regex=True)
+
+            # get and verify the file contents
+            self.user1.assert_icommand(f'iget -f {file1} {retrieved_file}')
```
alanking, line 2951: I'm wondering if it would be worth adding a test for these two cases (reading your existing tests made me curious):

1. put an object while resource is in consistent naming scheme; update naming scheme to decoupled; get the object
2. put an object while resource is in decoupled naming scheme; update naming scheme to consistent; get the object

Can it get the object? Do we expect it to be able to get the object? I honestly don't know.

alanking (commented):
Seems good overall

packaging/resource_suite_s3_nocache.py
trel: i think ... it would work?   writes would go into new paths, but gets... should just look em up, and go get em.. and they should still be there...

alanking: Would it? Updating the configuration on the resource actually updates the physical path for all of the replicas in the resource? It would effectively be the same as changing the vault path for a unixfilesystem resource which would require the admin to run `iadmin modrescdatapaths`, no?

alanking: And if that is the case, that probably falls outside the scope of this PR. Either way, I'm curious. :)

trel: hmm, no / i hope not?   changing a setting in the catalog is just a context string - the plugin wouldn't do anything until the next invocation, and at that point it doesn't know it was ever in a different mode (prior).

alanking: I created an issue here for this: https://github.com/irods/irods_resource_plugin_s3/issues/2161

packaging/resource_suite_s3_nocache.py
JustinKyleJames: As Alan said, this could be run in topology it would trigger a redirect.  When running on one host it will not trigger the redirect.

I added it just in case that functionality was available in the future.

s3/s3_operations.cpp
JustinKyleJames: Well, this didn't even work as I must have made a mistake in my manual testing.

Instead of using the open flags (which turned out to be zero), I am now checking the openType is CREATE_TYPE.  That seems to be working fine.  I verified the force overwrite, the openType is not CREATE_TYPE thus the physical_path is not updated meaning whatever was used before remains.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2995,6 +2999,10 @@ def test_resource_updated_from_decoupled_to_consistent(self):
             # update the resource to detached mode
             self.admin.run_icommand(f'iadmin modresc {resource_name} context {s3_context_consistent}')
 
+            # get and verify the file contents
+            self.user1.assert_icommand(f'iget -f {file1} {retrieved_file}')
+            self.assertTrue(filecmp.cmp(file1, retrieved_file))  # confirm retrieved is correct with file2
```
trel, line 3004: "correct with file2" -> "correct with file1"

packaging/resource_suite_s3_nocache.py
```diff
@@ -2940,6 +2940,10 @@ def test_resource_updated_from_consistent_to_decoupled(self):
             # update the resource to detached mode
             self.admin.run_icommand(f'iadmin modresc {resource_name} context {s3_context_decoupled}')
 
+            # get and verify the file contents
+            self.user1.assert_icommand(f'iget -f {file1} {retrieved_file}')
+            self.assertTrue(filecmp.cmp(file1, retrieved_file))  # confirm retrieved is correct with file2
```
trel, line 2945: "correct with file2" -> "correct with file1"

s3/s3_operations.cpp
```diff
@@ -401,22 +402,57 @@ namespace irods_s3 {
...
                 const auto s3_key_name = fmt::format("/{}/{}/{}", bucket_name, obj_id, object_name);
 
-                // update physical path
-                logger::debug("{}:{} ({}) [[{}]] updating physical_path to {}",
-                        __FILE__, __LINE__, __FUNCTION__, thread_id, s3_key_name.c_str());
-                object->physical_path(s3_key_name);
-                strncpy(L1desc[index].dataObjInfo->filePath, s3_key_name.c_str(), MAX_NAME_LEN);
-                L1desc[index].dataObjInfo->filePath[MAX_NAME_LEN - 1] = '\0';
+                // Update physical path but only on first creation otherwise the policy that was in effect
+                // at the time the object was first created rules.
```
trel, line 428: "created rules"...

does this mean, "created wins"?

s3/s3_operations.cpp
```diff
@@ -2231,6 +2267,14 @@ namespace irods_s3 {
 
     irods::error s3_notify_operation( irods::plugin_context& _ctx,
         const std::string* str ) {
+        if (is_cacheless_mode(_ctx.prop_map())) {
+
+            // Must update the physical_path in the L1desc[] table for decoupled naming.
+            // In the case of a redirect, this runs on the original connected server and this
+            // is the server that updates the database. This will open update for creates.
```
trel, line 2274: maybe this sentence can be clarified...
"This will open update for creates"

korydraughn:
At the moment we don't have an automated way of running these tests in a topology environment.

We need to make sure testing of this plugin in that type of environment is enforced following the merging of this PR.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2869,3 +2867,157 @@ def test_iput_large_file_with_checksum_issue_2124(self):
...
+
+            # get and verify the file contents
+            self.user1.assert_icommand(f'iget {file1} {retrieved_file}')
+            self.assertTrue(filecmp.cmp(file1, retrieved_file))  # confirm retrieved is correct
+
+            # verify in the physical path that decoupled mode was used
+            print(self.user1.run_icommand(['ils', '-l', file1])[0])   # just debug
+            self.user1.assert_icommand(f'ils -L {file1}', 'STDOUT_SINGLELINE', f'/{self.s3bucketname}/[0-9]+/{file1}', use_regex=True)
+
+            # update the resource to detached mode
```
alanking, line 2999: Should this be "consistent" instead of "detached"?

alanking (commented):
Good descriptive comments. I saw a couple of things in the comments for the test but overall this is solid.

packaging/resource_suite_s3_nocache.py

packaging/resource_suite_s3_nocache.py

s3/s3_operations.cpp
JustinKyleJames: I updated it to from "rules" to "wins".

s3/s3_operations.cpp
JustinKyleJames: The "open" should be "only". But I rewrote this to make it all more clear.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: Done

packaging/resource_suite_s3_nocache.py
```diff
@@ -2869,3 +2867,157 @@ def test_iput_large_file_with_checksum_issue_2124(self):
...
+
+            # get and verify the file contents
+            self.user1.assert_icommand(f'iget {file1} {retrieved_file}')
+            self.assertTrue(filecmp.cmp(file1, retrieved_file))  # confirm retrieved is correct
+
+            # verify in the physical path that consistent mode was used
+            print(self.user1.run_icommand(['ils', '-l', file1])[0])   # just debug
+            self.user1.assert_icommand(f'ils -L {file1}', 'STDOUT_SINGLELINE', f'/{self.s3bucketname}/{resource_name}/home/.+/{file1}', use_regex=True)
+
+            # update the resource to detached mode
```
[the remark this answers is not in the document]
JustinKyleJames: Done

trel:
clang-format and trailing whitespace

good good

trel:
ready for squashing, no pounds

trel:
issue numbers and content split look good to me.   let's wait on one more vote.

korydraughn (approved):
Agreed. Split looks good. Pound it.

JustinKyleJames:
> Agreed. Split looks good. Pound it.

Done

trel (approved):

Issue #2156, opened 2023-12-14
Non-deterministic build failure when building libs3


SwooshyCueb:
When building a debug build of the plugin post-#2152, the compiler sometimes crashes when building libs3 due to a bug in clang 13.0.0. The current workaround is to just retry the build until it succeeds. This bug does not occur when using a newer version of clang. We do not currently have plans to bump the version of clang that we provide.

trel:
Specifically... clang 13.0.1 appears to have fixed it.

We think it might be https://github.com/llvm/llvm-project/commit/67b5bc26bde81a60688ae5c049a001ffbc6dd614 (one of the [79 commits since 13.0.0](https://github.com/llvm/llvm-project/compare/llvmorg-13.0.0...llvmorg-13.0.1))

trel:
crash can be seen here... https://github.com/irods/irods_resource_plugin_s3/pull/2152#issuecomment-1856187018

trel:
oh, a twist!

trel:
and the crash is the 'same'?

SwooshyCueb:
Seems to be.

Here's the output and the dumped files from a crash I just observed:
https://gist.github.com/SwooshyCueb/d90275f4047fe5ed73c8a5c85eac233d

SwooshyCueb:
I wrote a script to compile the source files and headers over and over again until the compiler crashed, and found that only three files cause crashes:
- `bucket.c`
- `multipart.c`
- `request.c`

After some testing with `-Weverything`, I identified that these three files, and no others, would throw the `-Wvla` warning, and sure enough, if I swap the variable length array declarations for calls to `alloca`, the compiler crashes go away.

I'll submit a PR shortly.

trel:
Great stuff.

PR #2181, opened 2024-03-07
[#2156] Replace variable length arrays with `alloca()` (main)

SwooshyCueb:
Addresses #2156 (for real this time)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,9 @@ if (NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   message(STATUS "Setting unspecified CMAKE_BUILD_TYPE to '${CMAKE_BUILD_TYPE}'")
 endif()
 
+include(CheckCCompilerFlag)
+include(CheckCXXCompilerFlag)
+
 if (CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
   set(IRODS_BUILD_WITH_WERROR_DEFAULT ON)
 else()
@@ -45,6 +48,17 @@ add_compile_options(-Wall -Wextra)
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-fpermissive>)
+endif()
+
+# Variable length arrays can crash clang (#2156), so let's turn on warnings for them.
+# -Wvla is not enabled by -Wall or -Wextra, so we have to do it manually.
+check_c_compiler_flag(-Wvla can_use_wvla_c)
+check_cxx_compiler_flag(-Wvla can_use_wvla_cxx)
+if (can_use_wvla_c)
+  add_compile_options($<$<COMPILE_LANGUAGE:C>:-Wvla>)
+endif()
+if (can_use_wvla_cxx)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wvla>)
 endif()
 
 if (NOT DEFINED THREADS_PREFER_PTHREAD_FLAG)
--- a/libs3/src/bucket.c
+++ b/libs3/src/bucket.c
@@ -30,6 +30,7 @@
  *
  ************************************************************************** **/
 
+#include <alloca.h>
 #include <string.h>
 #include <stdlib.h>
 #include "libs3/libs3.h"
@@ -464,7 +465,7 @@ static S3Status make_list_bucket_callback(ListBucketData* lbData)
 	int isTruncated = (!strcmp(lbData->isTruncated, "true") || !strcmp(lbData->isTruncated, "1")) ? 1 : 0;
 
 	// Convert the contents
-	S3ListBucketContent contents[lbData->contentsCount];
+	S3ListBucketContent* contents = alloca(sizeof(S3ListBucketContent) * lbData->contentsCount);
 
 	int contentsCount = lbData->contentsCount;
 	for (i = 0; i < contentsCount; i++) {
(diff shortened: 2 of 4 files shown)


korydraughn (commented):
The changes look fine, but I'm not sold on the use of `alloca` yet.

I don't have any experience with it, but reading about it indicates we should probably rethink using it.

The main thing is that we're replacing VLAs with `alloca`, which seems incorrect.

Thoughts?

SwooshyCueb:
I'm not sure I understand the problem. `alloca()` allocates on the stack, just as a VLA does.

korydraughn:
Right. I'm concerned about stack overflow and undefined behavior.

From `alloca`'s man page (https://man7.org/linux/man-pages/man3/alloca.3.html):
```
The alloca() function returns a pointer to the beginning of the
allocated space.  If the allocation causes stack overflow,
program behavior is undefined.
```

The amount of stack space needed is unknown for the some/all of the VLAs.

Here's another link on it: https://stackoverflow.com/a/1018865/5818611

Another thing that comes to mind is if something weird happens as a result of using `alloca`, will we be able to track down the cause?

korydraughn:
The **NOTES** section of the man page makes me slightly nervous too.

SwooshyCueb:
Looking at the comments in the stackoverflow answer you linked, it appears that VLAs have the exact same problems as `alloca()` in regards to stack overflows.

What do you find concerning in the NOTES section?

korydraughn:
That section states its use is discouraged. I take that as the dev must be sure that function is the correct choice for the task and it doesn't leave room for stack overflow.

SwooshyCueb:
If my understanding is correct, given our usage, if a stack overflow would occur with `alloca()`, it would occur with VLAs as well. Therefore, if we have a potential stack overflow issue, we had it prior to this PR. The case can certainly be made that we should not be allocating space on the stack at runtime like this in the first place, but that's outside the scope of this PR.

korydraughn:
That's a fair point.

Please run the test suite to confirm everything is still good.

korydraughn:
That's unexpected.

Was anything reported that indicates the reason for the failures?
Does it fail at the same place / in the same way on every run?

SwooshyCueb:
So I forgot that the testing environment will report that all tests fail for plugin tests regardless of how many actually fail, and it turns out that on tip of main, only *most* tests fail. I think it may have something to do with irods/irods#7502. I'll open an issue about that.

For this PR, all tests did actually fail, and it appeared to be due to a signature mismatch. I figured out that this was due to a difference in how `sizeof()` works with `alloca()` vs VLAs, so I've replaced the relevant instances of `sizeof()` (see the new `[SQUASH]` commit), and now the test failures match tip of main.

I don't like clang-format's suggestion for the `canonicalRequestLen` definition, so I'm ignoring it.

korydraughn:
> and now the test failures match tip of main.

Are the failing tests you're referring to part of the irods/irods test suite or this repo's test suite?

> I don't like clang-format's suggestion for the canonicalRequestLen definition, so I'm ignoring it.

Agreed. Wrap it in `// clang-format off/on` so future edits don't yell at us about it.

SwooshyCueb:
> Are the failing tests you're referring to part of the irods/irods test suite or this repo's test suite?

They're all from this repo

SwooshyCueb:
I don't think so. I think they're failing because they weren't updated to expect the deprecation message introduced in irods/irods#7502. I'm still looking into it. There may be other tests failing for different reasons.

SwooshyCueb:
Changing the tests to expect the deprecation message fixed some of them, but another one, `test_irods_resource_plugin_s3_minio.Test_Compound_With_S3_Resource.test_irepl_multithreaded` fails for what seems to be an unrelated reason, and leaves the server in a state that causes every subsequent test to fail. This test is inherited from `ResourceBase` in the irods repo.

SwooshyCueb:
I've opened #2184 to cover the remaining test failures. The cause of the failure wasn't evident to me so I'm going to leave that work to someone else so I can get back to working on buildsystem and packaging stuff.

packaging/test_irods_resource_plugin_s3_for_cloudian.py
```diff
@@ -328,7 +328,9 @@ def test_irepl_with_purgec(self):
         # assertions
         self.admin.assert_icommand_fail("ils -L "+filename,'STDOUT_SINGLELINE',filename) # should not be listed
         self.admin.assert_icommand("iput "+filename) # put file
-        self.admin.assert_icommand("irepl -R "+self.testresc+" --purgec "+filename) # replicate to test resource
+        self.admin.assert_icommand(
+        "irepl -R " + self.testresc + " --purgec " + filename,
+        'STDOUT', 'Specifying a minimum number of replicas to keep is deprecated.') # replicate to test resource
```
alanking, line 333: Funky indentation here

alanking (commented):
I think I'm okay with the changes since `alloca` presents nothing more dangerous than VLAs. I just saw one thing in the tests.

Deferring to @korydraughn, though.

korydraughn:
Let's hold on pounding. We need to determine why the tests are failing. Once that's handled we can proceed.

SwooshyCueb:
We merged the test and include fixes in a separate PR (#2189,#2190) so they could make it in for 4.3.2. They've been removed from this PR (and the 4-3-stable PR).

trel:
Going to wait until after 4.3.2 to merge this PR.

SwooshyCueb:
I've confirmed that tests pass now. If there are no objections, I will \# this PR and the 4-3-stable PR.

korydraughn (approved):
Pound it.

SwooshyCueb:
\#'d

Issue #2180, opened 2024-03-07
`request_context.h` is missing `#include <curl/curl.h>`


SwooshyCueb:
`request_context.h` references `CURLM` but does not `#include` any curl headers:
https://github.com/irods/irods_resource_plugin_s3/blob/015b149873fbf402a3b4adb038349b6a2db97bc0/libs3/include/libs3/request_context.h#L46

PR #2182, opened 2024-03-07
[#2156,#2180] Replace variable length arrays with `alloca()`, and add missing curl include to `request_context.h` (4-3-stable)

SwooshyCueb:
Addresses #2156 (for real this time)
Addresses #2180


SwooshyCueb:
\#'d

korydraughn:
4-3-stable relies on libc++. Was main tested using libstdc++ or libc++?

I noticed the comment about VLAs crashing clang in the cmake. Is that something we need to worry about for this branch?

SwooshyCueb:
> 4-3-stable relies on libc++. Was main tested using libstdc++ or libc++?

I tested with libstdc++ on main, libc++ on stable.

> I noticed the comment about VLAs crashing clang in the cmake. Is that something we need to worry about for this branch?

There's no difference between the two branches that affects clang's tendency to crash when compiling source code that contains VLAs.

PR #2199, opened 2024-06-03
[2198] Do not blindly set #threads to requested #threads (4-2-stable)

JustinKyleJames:
Test run of the main suite of tests passed.

PR #2201, opened 2024-06-04
Issue 2198 43 stable

JustinKyleJames:
This is the same as pull request #2199 but with Python3 changes as well as a slight change in stdout when iput -r is run in 4.3.


trel:
so this was.... targeted incorrectly?

PR #2217, opened 2024-09-25
[2216] Updated README.md for troubleshooting checksum read request rate errors (4-3-stable)

JustinKyleJames:
This change must be applied to 4.3.4.0 (not 4.3.3.1) since the new setting is not valid without the corresponding updates to iRODS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,23 @@ To increase or decrease the logging level for the S3 plugin, add an `s3_resource
 },
 ```
 
+## Troubleshooting
+
+### Handling "Reduce Your Request Rate" Errors
+
+On operations that require calculating a checksum for an S3 object at rest, iRODS must read the object's data to calculate the checksum. The default buffer size for these reads is 1 MiB. Each of these read requests map to a separate GetObject request to S3 with a range of bytes. This can exceed the request rate that the S3 provider allows.
+
+To resolve this problem, the read buffer size can be increased by updating the `checksum_read_buffer_size_in_bytes` value in the `advanced_settings` stanza of server_config.json. This change should be made for all servers that can serve requests to that S3 resource.
+
+The following is an example of increasing this buffer size to 10 MiB:
+
+```
+    "advanced_settings": {
+        ...
+        "checksum_read_buffer_size_in_bytes": 10485760
+        ...
+```
+
 ## Using this plugin with Google Cloud Storage (GCS)
 
 This plugin has been manually tested to work with Google Cloud Storage, with some caveats.


README.md
```diff
@@ -229,6 +229,23 @@ To increase or decrease the logging level for the S3 plugin, add an `s3_resource
 },
 ```
 
+## Troubleshooting
+
+### Handling Reduce Your Request Rate Errors
+
+On operations that require a checksum calculation to an S3 object at rest, iRODS must read the object's data to calculate the checksum. The default buffer size for these reads is 1 MiB. Each of these read requests map to a separate GetObject request to S3 with a range of bytes. This can exceed the request rate that the S3 provider allows.
```
alanking, line 236:
```suggestion
On operations that require calculating a checksum calculation for an S3 object at rest, iRODS must read the object's data to calculate the checksum. The default buffer size for these reads is 1 MiB. Each of these read requests map to a separate GetObject request to S3 with a range of bytes. This can exceed the request rate that the S3 provider allows.
```

README.md
```diff
@@ -229,6 +229,23 @@ To increase or decrease the logging level for the S3 plugin, add an `s3_resource
...
+
+To resolve this problem, the read buffer size can be increased by updating the `checksum_read_buffer_size_in_bytes` value in the `advanced_settings` stanza of server_config.json.
```
alanking, line 238: Maybe mention that the setting needs to be updated on all servers which can service the S3 resource requests? Maybe that's implied, but could be good to say out loud (especially for the "detached" use cases).

JustinKyleJames: Good point.

README.md
JustinKyleJames: Updated as listed above but left out the "calculation" after checksum.

README.md
JustinKyleJames: done

README.md
alanking: Oh, I didn't read the whole sentence. My bad...

alanking (commented):
Looks good to me. Let's get another approval before we merge. Also, I suppose we should actually merge the feature before this, too. Also-also, would we have to wait for iRODS 4.3.4 and 4.3.4.0 of this plugin before merging this, since it wouldn't be true until both were released? Hmm...

README.md
```diff
@@ -229,6 +229,23 @@ To increase or decrease the logging level for the S3 plugin, add an `s3_resource
 },
 ```
 
+## Troubleshooting
+
+### Handling Reduce Your Request Rate Errors
```
korydraughn, line 234: Wrap `Reduce Your Request Rate` in double quotes so that phrase appears as a single unit.

trel (approved):

korydraughn (commented):
> Also, I suppose we should actually merge the feature before this, too.

Agreed.

> Also-also, would we have to wait for iRODS 4.3.4 and 4.3.4.0 of this plugin before merging this, since it wouldn't be true until both were released? Hmm...

That's correct. Let's get this PR to a merge-able point and leave it open until we know what iRODS 4.3.4 looks like. Once this PR is ready, we cherry-pick and merge it into the main branch.

So ...
1. Polish this PR for approval, but do not merge.
2. Cherry-pick and merge into the main branch.

README.md
JustinKyleJames: done

korydraughn:
Looks like this can be squashed (no pounds) and cherry-picked to 4-3-stable.

JustinKyleJames:
> Looks like this can be squashed (no pounds) and cherry-picked to 4-3-stable.

I think you meant to main.  I squashed and cherry-picked to main.  Will open a PR for it.

korydraughn:
Yes, that's what I meant.

JustinKyleJames:
> Yes, that's what I meant.

Is it ready for pounding?

alanking:
The plan is to cherry-pick this to main and merge it there but hold on this one until we're closer to releasing 4.3.4. See https://github.com/irods/irods_resource_plugin_s3/pull/2217#pullrequestreview-2328916073. The idea - I think - is that we don't want to advertise the feature when the server version supplying said feature has not yet been released. Alternatively, we can merge now and hope that users view the README for the specific release tag they are using.

korydraughn:
What @alanking has explained is correct.

Please cherry-pick this work to the main branch and we'll merge that. We'll keep this PR open until we're closer to a 4.3.4 release.

korydraughn:
I think we're ready to look at getting this merged now that 4.3.4 is close.

@JustinKyleJames Please test this PR to make sure it is still in a good spot.

korydraughn:
I just realized this is just documentation. Ignore my comment about running the tests.

Will review the words once more before approving.

korydraughn (approved):
The wording in this PR is identical to what's in the main branch.

Pound it.

alanking (approved):

Issue #2225, opened 2025-03-01
Test asserts incorrect error code against iRODS 4.3.4


korydraughn:
- [x] main
- [x] 4-3-stable
---
## Testing

The following test/assertion fails due to the `iput` returning `USER_SOCK_CONNECT_TIMEDOUT`.

https://github.com/irods/irods_resource_plugin_s3/blob/a2c8319903e5596680ac71aa657b67f21d76e32f/packaging/resource_suite_s3_nocache.py#L1562

Updating this line allows all tests to pass.

Please investigate why a different error code is being returned.

korydraughn:
Adding a few log statements to the iRODS server points to this location being responsible for the error code.

https://github.com/irods/irods/blob/3f6524069c0eac458fc26808b7095a414581a28e/lib/core/src/sockComm.cpp#L1030-L1035

That code hasn't changed in significant ways to cause this. There must be another place which explains this change in behavior.

korydraughn:
Looking at the test history for each 4.3 release, everything shows this test passing and seems to imply that something has changed between 4.3.3 and 4.3.4. However, I compiled and ran the 4.3.3 version of the server and resource plugin and was able to reproduce this test failure consistently 100% of the time. So, it's not clear to me why we didn't see this same behavior during release testing of 4.3.3 and earlier.

To add to that, there's been only 2 commits added to this repo since 4.3.3 and neither of them change the implementation of the plugin.

All of this tells me that the change which made this test failure more consistent occurred in the iRODS server.

Overall, I don't think this is a blocker because the change in error code isn't related to the resource plugin. It is reproducible with other resource plugins. Using 4.3.3, I was able to reproduce this situation using `iput` and a unixfilesystem resource. So, I think this test is more of a server test rather than a plugin test - i.e. it's independent of any plugin because it has to do with establishing socket connections between iRODS servers and that normally happens before hitting any plugin logic.

trel:
that's good and all sounds consistent.    agreed, just get the test to pass for this plugin and I think we're good for now.

korydraughn:
Came across irods/irods#7990 while preparing 4.3.4 release notes.

This indicates that the change in behavior could be due to firewall/network changes. Reopening this to update the test so that it accepts both error codes.

PR #2226, opened 2025-03-01
[#2225] Change expected error code in test (4-3-stable)

korydraughn:
This change to the test allows a full test run to pass against iRODS 4.3.4.

Still need to determine what caused a change in the error code.


korydraughn:
See issue #2225 for reasons this is safe to merge.

Merging.

PR #2237, opened 2025-03-31
[irods/irods#8287] Replace gethostname() with "host" server config option (main)

korydraughn:
Will confirm it builds and I think we can move forward with merging this.

If there are any issues, we can address them after the server is wrapped up.


alanking (dismissed):
Seems fine to me. I'm assuming that if the host property is not correct, the server won't even be able to function, so shouldn't really need to worry about that part at this point in the code.

\# it once confirmed building.

korydraughn:
Yes.

alanking (approved):
Okay, I'll keep an eye out.

\# it, if building

korydraughn:
Confirmed that changes build.

Added pounds.

PR #2242, opened 2025-04-09
[2241] Use file_open_counter to determine open mode if number of threads is unknown (4-3-stable)

JustinKyleJames:
This is to fix opens with truncate flag when another thread already opened the cache file.

PR #2244, opened 2025-04-14
[2241] Use file_open_counter to determine open mode if number of thre… (main)

JustinKyleJames:
This is a cherry-pick from #2242.


JustinKyleJames:
> Looks identical. # it

done

PR #2272, opened 2026-01-07
[irods/irods#8502] Prevent CREATE op for archive mode (main)

alanking:
Addresses irods/irods#8502

All S3 tests passed.

Interestingly, only "Prevent CREATE op for archive mode" is required to resolve the issue. However, we may wish to tighten up the accepted context string space anyway, so I kept that change in a separate commit.

### DIFF
--- a/s3_resource/src/s3_operations.cpp
+++ b/s3_resource/src/s3_operations.cpp
@@ -2356,6 +2356,13 @@ namespace irods_s3 {
         logger::debug("{}:{} ({}) [[{}]] get_resource_name={}",
                 __FILE__, __LINE__, __FUNCTION__, thread_id, irods::get_resource_name(_ctx).c_str());
         _out_parser->add_child(irods::get_resource_name(_ctx));
+
+        // Create operations are not allowed for archives. Any data put into an archive will come through the
+        // "sync_to_arch" operation.
+        if (!is_cacheless_mode(_ctx.prop_map()) && irods::CREATE_OPERATION == *_opr) {
+            return ERROR(SYS_NOT_SUPPORTED, "Create operation not supported for an archive");
+        }
+
         *_out_vote = irv::vote::zero;
         try {
             *_out_vote = irv::calculate(*_opr, _ctx, *_curr_host, *_out_parser);


packaging/resource_suite_s3_nocache.py
```diff
@@ -2357,6 +2357,77 @@ def test_client_side_and_server_side_verify_crc64nvme_checksums_on_put_operation
...
+
+        # Resouce and data for test.
+        s3_resc = "demoResc"
+        data_object_name = "muhstuff"
+        logical_path = "/".join([self.user0.session_collection, data_object_name])
+        content = "muhtext"
+
+        try:
+            # Update the resource context string with the bad HOST_NAME configuration.
+            new_context = re.sub(right_host_mode, wrong_host_mode, original_context)
```
korydraughn: Let's assert that `new_context` isn't equal to `original_context`.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2357,6 +2357,77 @@ def test_client_side_and_server_side_verify_crc64nvme_checksums_on_put_operation
...
+        data_object_name = "muhstuff"
+        logical_path = "/".join([self.user0.session_collection, data_object_name])
+        content = "muhtext"
+
+        # archive_attached is the default HOST_MODE if no HOST_MODE is specified, or it can be set explicitly.
+        for new_host_mode in ["", "HOST_MODE=archive_attached"]:
+            with self.subTest(new_host_mode):
+                try:
+                    # Update the resource context string with the archive_attached HOST_NAME configuration.
+                    new_context = re.sub(original_host_mode, new_host_mode, original_context)
```
korydraughn: Let's assert that `new_context` isn't equal to `original_context`.

packaging/resource_suite_s3_nocache.py
alanking: Done

alanking:
> Are you saying that by immediately returning an error, the plugin's state becomes unknown? Therefore, we can't be sure how the plugin will behave afterwards?
>
> Please say more about this.

No, it's just that the implementation that I had was trying to minimize the number of changes and would throw an `irods::exception` if an invalid configuration was detected. I added a couple of catch blocks for these but then found that the function I modified is called from another function that is called in nearly every plugin operation. What this means is that I would have to go through and wrap every call with a try-catch block to make sure that the exceptions are handled. The implementation as it was would have *not* caught those exceptions, which means that the plugin operations could throw.

I call it a violation of the plugin interface because the resource plugin operations are not supposed to throw (as far as I'm aware), but should return an `irods::error`. In my view, a better fix would be to modify the return value of the internal function that checks the `HOST_MODE`.

An even better fix would probably be catching it at modification time, or in a new plugin operation, as you suggested. And also here: https://github.com/irods/irods/issues/8502#issuecomment-3032879176.

> Is there anything else left to do for this PR?

I'm going to run the full test suite again with the latest changes just in case. The new tests are passing so I have almost-full confidence that it won't be a problem, but I'd like to see the success for myself

korydraughn:
I agree with that.

packaging/resource_suite_s3_nocache.py
```diff
@@ -2357,6 +2357,79 @@ def test_client_side_and_server_side_verify_crc64nvme_checksums_on_put_operation
...
+                    self.assertNotEqual(new_context, original_context)
+                    self.admin.assert_icommand(["iadmin", "modresc", s3_resc, "context", new_context])
+
+                    # Try to write to logical_path using the S3 misconfigured S3 resource.
```
alanking:
```suggestion
                    # Try to write to logical_path using the S3 resource.
```

korydraughn (approved):
Pound it.

PR #2304, opened 2026-07-22
[2256] Moved g_mpuData free to at_scope_exit

JustinKyleJames:
Also moved manager.xml, manager.upload_id, and manager.etags[i] frees to the same.

Tests passed.

### DIFF
--- a/s3_resource/src/s3_resource.cpp
+++ b/s3_resource/src/s3_resource.cpp
@@ -1888,13 +1888,25 @@ irods::error s3PutCopyFile(
             return ERROR( SYS_MALLOC_ERR, msg );
         }
         g_mpuData = new multipart_data_t[totalSeq];
+        const auto free_mpu_data = irods::at_scope_exit{[&manager, totalSeq] {
+            delete[] g_mpuData;
+            g_mpuData = nullptr;
+            free(manager.xml);
+            free(manager.upload_id);
+            if (manager.etags) {
+                for (int i = 0; i < totalSeq; i++) {
+                    free(manager.etags[i]);
+                }
+                free(manager.etags);
+            }
+        }};
+
         if (!g_mpuData) {
             // Clear up the S3PutProperties, if it exists
             if (putProps) {
                 if (putProps->md5) free( (char*)putProps->md5 );
                 free( putProps );
             }
-            free(manager.etags);
             const auto msg = fmt::format("[resource_name={}] Out of memory error in S3 multipart g_mpuData allocation.", resource_name);
             s3_logger::error( msg );
             return ERROR( SYS_MALLOC_ERR, msg );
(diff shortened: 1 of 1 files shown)


s3_resource/src/s3_resource.cpp
```diff
@@ -1888,13 +1888,31 @@ irods::error s3PutCopyFile(
             return ERROR( SYS_MALLOC_ERR, msg );
         }
         g_mpuData = new multipart_data_t[totalSeq];
+        const auto free_mpu_data = irods::at_scope_exit{[&manager, totalSeq] {
+            delete[] g_mpuData;
+            g_mpuData = nullptr;
+            if (manager.xml) {
```
FifthPotato: Do we need to check this before `free()`ing? If I recall correctly, you can pass null pointers to `free()` without issue. (Also applies to the other `free()`s without element accesses.)

FifthPotato (commented):
Changes seem reasonable to me. Mind the typo in the commit message and the PR title `at_scope_exist` -> `at_scope_exit`

korydraughn:
GitHub Actions is reporting test failures. Please take a look at those.

s3_resource/src/s3_resource.cpp
JustinKyleJames: If you are referring to the delete[], that is safe to do without checking for a null pointer.

If you are referring to the check before free(), that isn't technically necessary anymore either.  (It was at one point.)

I will remove those checks.

JustinKyleJames: done

JustinKyleJames:
> Changes seem reasonable to me. Mind the typo in the commit message and the PR title `at_scope_exist` -> `at_scope_exit`

I fixed the typo.

s3_resource/src/s3_resource.cpp
```diff
@@ -1954,6 +1963,10 @@ irods::error s3PutCopyFile(
         if (_mode == S3_COPYOBJECT) {
             ret = parseS3Path(_filename, srcBucket, srcKey, _prop_map);
             if (!ret.ok()) {
+                if (putProps) {
+                    if (putProps->md5) free( (char*)putProps->md5 );
+                    free( putProps );
+                }
```
korydraughn, line 1969: Should `putProps` be set to `nullptr`? Is it reused anywhere?

s3_resource/src/s3_resource.cpp
trel, line 2091: i notice the same variable here... is this okay as is?

s3_resource/src/s3_resource.cpp
JustinKyleJames: All of the frees for `putProps` in this routine have immediate returns afterward.  It is not used afterward before the returns.  It is also a local variable.

s3_resource/src/s3_resource.cpp
JustinKyleJames: Yes, `putProps` is a local variable and the return happens right after this.  (Same with all of the others in this routine.)

korydraughn (commented):
Squash to taste.

JustinKyleJames:
> Squash to taste.

done

korydraughn (approved):
Pound it.

JustinKyleJames:
> Pound it.

Done

korydraughn:
Merging.

PR #2312, opened 2026-08-10
[2185] Do not drain circular buffer if object size requires > 10000 …

JustinKyleJames:
…parts.

The normal part size is the circular_buffer_size. This is to allow retries on part upload failures. However, for extremely large size objects this can result in an attempt to upload more than 10,000 parts which is the S3 limit.

To resolve this do the following:

1. Upon open of a multipart write, determine the part size:
    - In most cases this will be the circular buffer size.
    - For very large files this will be roughly object_size / 10,000.
2. Added a new transport instance variable (part_size_) which defaults to the circular_buffer_size. Use this rather than the circular_buffer_size throughout the code for the part size.
3. Added a new instance variable (draining_enabled_) in the transport class to store whether draining the circular buffer must be drained on reading or only read without draining. The default false to keep the data in the buffer so retries can be performed on failure.
4. When bullet 2 is encountered, calculate part_size_  and set draining_enabled_ to true.
5. When draining_enabled_ is true, remove bytes from the circular buffer upon reading and do not attempt retries on part upload failures.

Also added unit tests to cover the math and to simulate the condition where a large file would force draining_enabled_ to true.


trel:
commit message itself is hard to parse.   i think there are a couple typos.

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,185 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+    {
+        std::int64_t part_size = -1;
+        bool draining_required = true;
+
+        bool ok = s3_transport::compute_part_size_for_object(
+                /* object_size */                     50*1024*1024,
+                /* circular_buffer_size */              10*1024*1024,
+                /* number_of_client_transfer_threads */ 4,
+                /* max_number_of_parts */               10,
+                /* max_part_size */                     1024*1024*1024,
```
trel: do we normally put the comment after / to the right of the code?

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,185 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+    {
+        std::int64_t part_size = -1;
+        bool draining_required = false;
+
+        bool ok = s3_transport::compute_part_size_for_object(
+                /* object_size */                      1000000,
+                /* circular_buffer_size */              10,
+                /* number_of_client_transfer_threads */ 1,
+                /* max_number_of_parts */               10,
+                /* max_part_size */                     1000,
```
trel: same

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,185 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+    {
+        std::int64_t part_size = -1;
+        bool draining_required = true;
+
+        bool ok = s3_transport::compute_part_size_for_object(
+                /* object_size */                      s3_transport_config::UNKNOWN_OBJECT_SIZE,
+                /* circular_buffer_size */              10*1024*1024,
+                /* number_of_client_transfer_threads */ 4,
+                /* max_number_of_parts */               10000,
+                /* max_part_size */                     5LL*1024*1024*1024,
```
trel: same

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -804,35 +808,101 @@ namespace irods::experimental::io::s3_transport
...
+        //     5 GiB in S3 standards).
+        //
+        //   - false if the calculated part_size is greater than max_part_size.
+        //
+        //     This can happen due to the following scenarios:
+        //     - file_size > max_part_size * safe_max_parts - With the default values
+        //       this is 50 TiB which is the largest object possible in S3.
+        //     - circular_buffer_size > max_part_size.  This is a misconfiguration and is unlikely
+        //       as it means each buffer is greater than 5 GiB.
+        static bool compute_part_size_for_object(
```
FifthPotato: Do we agree with this returning `bool`? It looks like it's returning a comparison of `part_size` against one of its inputs. If the caller gets access to `part_size` the caller could derive the value of `part_size <= max_part_size` themselves.
Alternatively, maybe the name could be updated to represent everything this function is doing? (Calculating if `part_size` is OK, returning `part_size`, returning `draining_required`, etc.)

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -804,35 +808,101 @@ namespace irods::experimental::io::s3_transport
...
+                std::int64_t object_size,
+                std::uint64_t circular_buffer_size,
+                int number_of_client_transfer_threads,
+                std::int64_t max_number_of_parts,
+                std::int64_t max_part_size,
+                std::int64_t& part_size,
+                bool& draining_required)
```
FifthPotato: Is there a way we could avoid the ref-pass as return value? It's probably fine as-is, but based on the root comment, it seems like `draining_required` will only ever be true if, after all the calculations are done, `part_size != circular_buffer_size`*. That means if the caller has access to `part_size`, they could calculate the value of `draining_required` themselves instead of making this function do it.

\*`!=` here for correctness; in non-pathological cases, it should be `part_size > circular_buffer_size`

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -804,35 +808,101 @@ namespace irods::experimental::io::s3_transport
...
                 : file_offset / thread_number;
 
-            // Determine number of parts per thread.  If parts is not divisible by circular buffer
-            // size then we need one additional part. The last thread is treated specially because
+            // Determine number of parts per thread.  If parts is not divisible by part_size
+            // then we need one additional part. The last thread is treated specially because
             // it may have additional bytes.
-            unsigned int parts_per_thread = bytes_all_threads_except_last / circular_buffer_size
-                + ( bytes_all_threads_except_last % circular_buffer_size == 0 ? 0 : 1 );
+            unsigned int parts_per_thread = bytes_all_threads_except_last / part_size
```
FifthPotato: This is just dividing and rounding up, I think. Since we do a different way of rounding up elsewhere (e.g. `std::int64_t parts_at_current_size = (object_size + part_size - 1) / part_size;` below), should we pick one and stick with it for consistency?
From what I can tell, the only potential downside would be maybe hitting an overflow if we choose the other way, but I'm not sure if that's something we need to be worried about. Thoughts?

(Applies to lines 821-822 below as well)

FifthPotato (commented):
Pretty involved PR! Nothing looks obviously incorrect, but I have limited knowledge of this plugin in general. Just going to offer some stylistic comments/questions.

JustinKyleJames:
> commit message itself is hard to parse. i think there are a couple typos.

I read through it and made changes.  Please check again.

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: I changed it to use `//` style trailing comments.

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: I will take a look at this.  I also need to fix the asserts because I don't think we should be asserting things which would cause a crash.  I need to roll up errors instead.

JustinKyleJames: None of the asserts are new so I am not going to touch those.

trel:
yes, comments/commit messages look good to me.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: I changed this to return an irods::error instead of a bool.  I think that makes it easier to understand that the one condition is an error condition and when that is encountered that error gets stored for later.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: I changed `validate_and_compute_part_size` to `validate_and_compute_part_size_and_set_buffer_draining_flag`.  I moved the draining logic to this wrapper function and removed it as a variable passed by reference.

The part size is still passed by reference and set in `compute_part_size_for_object`.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: These are equivalent ways to round up.  I'll make the change but keep in mind these lines were unchanged except for `circular_buffer_size` being replaced by `part_size` so that might confuse the purpose of the change a bit.

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+    Test_S3_NoCache_Draining in test_irods_resource_plugin_s3.py, which is skipped by
+    default and must be explicitly re-enabled to run this.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """Set up the test."""
+        if not hasattr(self, 'proto'):
+            self.proto = 'HTTPS'
+
+        super(Test_S3_NoCache_Draining_Base, self).__init__(*args, **kwargs)
```
korydraughn, line 3467: Why is this needed? I don't think I've seen any tests use `__init__`.

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+
+    def setUp(self):
+
```
korydraughn: Remove empty line?

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+        self.s3_context += ';S3_MPU_CHUNK=5'
+        self.s3_context += ';CIRCULAR_BUFFER_SIZE=2'
+        self.s3_context += ';S3_CACHE_DIR=/var/lib/irods'
+
+        self.admin.assert_icommand(
+            ['iadmin', 'mkresc', self.testresc, 's3', hostname + ":/" + self.s3bucketname + self.testvault, self.s3_context],
+            'STDOUT_SINGLELINE', 's3')
+
+    def tearDown(self):
+
```
korydraughn: Remove empty line?

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+            self.user0.assert_icommand("iget -f %s %s" % (file1, file2))
+
+            file2_hash = s3plugin_lib.sha256sum(file2)
+            self.assertEqual(file1_hash, file2_hash)
+
+            self.user0.assert_icommand("irm -f %s" % file1, 'EMPTY')
+
+        finally:
+
```
korydraughn: Remove empty line?

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+        if not hasattr(self, 'proto'):
+            self.proto = 'HTTPS'
+
+        super(Test_S3_NoCache_Draining_Base, self).__init__(*args, **kwargs)
+
+    def setUp(self):
+
+        super(Test_S3_NoCache_Draining_Base, self).setUp()
+        self.admin = self.admin_sessions[0]
+        self.user0 = self.user_sessions[0]
```
korydraughn, line 3472: No `self.user1` for the bobby user? Is that intentional?

Is the second rodsuser needed?

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+
+        hostname = lib.get_hostname()
+
+        # read aws keys
+        self.read_aws_keys()
```
korydraughn: Comment can be removed.

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+        self.s3_context += ';S3_AUTH_FILE=' + self.keypairfile
+        self.s3_context += ';S3_REGIONNAME=' + self.s3region
+        self.s3_context += ';S3_RETRY_COUNT=2'
+        self.s3_context += ';S3_WAIT_TIME_SECONDS=3'
+        self.s3_context += ';S3_PROTO=' + self.proto
+        self.s3_context += ';HOST_MODE=cacheless_attached'
+        self.s3_context += ';S3_ENABLE_MPU=1'
+        self.s3_context += ';S3_MPU_CHUNK=5'
+        self.s3_context += ';CIRCULAR_BUFFER_SIZE=2'
+        self.s3_context += ';S3_CACHE_DIR=/var/lib/irods'
```
korydraughn, line 3510: Consider creating a dedicated directory under `/var/lib/irods` or `/tmp` instead.

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+                region=self.s3region,
+                secure=(self.proto == 'HTTPS'))
+
+        if hasattr(self, 'static_bucket_name'):
+            self.s3bucketname = self.static_bucket_name
+        else:
+            objects = s3_client.list_objects(self.s3bucketname, recursive=True)
+            for obj in objects:
+                s3_client.remove_object(self.s3bucketname, obj.object_name)
+            s3_client.remove_bucket(self.s3bucketname)
```
korydraughn, line 3534: Should the removal of the bucket be moved out of the else-block?

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+
+    def read_aws_keys(self):
```
korydraughn, line 3536: Consider renaming this function so it's clear that it creates new member variables on the instance e.g. `load_aws_keys_into_member_variables`. That's not a great name, but is pretty clear.

packaging/resource_suite_s3_nocache.py
```diff
@@ -3437,3 +3437,149 @@ def test_resource_updated_from_decoupled_to_consistent_issue_2161(self):
...
+        with open(self.keypairfile) as f:
+            self.aws_access_key_id = f.readline().rstrip()
+            self.aws_secret_access_key = f.readline().rstrip()
+
+    def test_upload_download_object_that_requires_circular_buffer_draining__issue_2185(self):
+        """Uploads a real object large enough to force part_size beyond the configured
+        circular_buffer_size (10MB here), engaging draining_enabled_ in the actual
+        plugin -- not just in the C++ unit tests, which bypass real configuration
+        validation and so can't reach this threshold with a small file.  See issue
+        #2185 and s3_transport::compute_part_size_for_object.
```
korydraughn: We need to avoid mentioning member variables in tests since names can change. Is there another way to say this without mentioning variables in the plugin's implementation?

This applies to `draining_enabled_` and `s3_transport::compute_part_size_for_object`.

packaging/test_irods_resource_plugin_s3.py
```diff
@@ -178,6 +179,17 @@ def __init__(self, *args, **kwargs):
         self.s3stsdate=''
         super(Test_S3_Cache_Glacier, self).__init__(*args, **kwargs)
 
+@unittest.skip('issue #2185 draining test requires ~100GB free disk, a long run time, and incurs '
+                'a small real cost against AWS -- remove this skip to run it manually')
```
korydraughn: Please introduce an external control for this, just like the one in the class below.

packaging/test_irods_resource_plugin_s3_minio.py
```diff
@@ -139,6 +140,19 @@ def __init__(self, *args, **kwargs):
         self.s3EnableMPU=1
         super(Test_S3_NoCache_EU_Central_1, self).__init__(*args, **kwargs)
 
+@unittest.skip('issue #2185 draining test requires ~100GB free disk on the client side plus '
+                '~100GB more for the local MinIO backing store (same host in this test setup), '
+                'and a long run time -- remove this skip to run it manually')
```
korydraughn: Please introduce an external control for this, just like for the CRC64/NVME tests.

s3_transport/include/irods/private/s3_transport/callbacks.hpp
```diff
@@ -841,7 +842,15 @@ namespace irods::experimental::io::s3_transport
                         : this->content_length - this->bytes_written;
 
                     try {
-                        circular_buffer.peek(this->bytes_written, bytes_to_return, libs3_buffer);
+                        // When draining_enabled, the buffer is drained (destructively read) as we
```
korydraughn:
```suggestion
                        // When draining_enabled is true, the buffer is drained (destructively read) as we
```

s3_transport/include/irods/private/s3_transport/callbacks.hpp
```diff
@@ -884,6 +893,12 @@ namespace irods::experimental::io::s3_transport
                 }
 
                 void post_success_cleanup() {
+                    // If draining_enabled, bytes were already removed from the buffer as they
```
korydraughn:
```suggestion
                    // If draining_enabled is true, bytes were already removed from the buffer as they
```

korydraughn (commented):
I've looked at the majority of the files. Will review the ones I did not get to later.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -804,35 +807,105 @@ namespace irods::experimental::io::s3_transport
...
+        // Note that in determine_start_and_end_part_from_offset_and_bytes_this_thread,
+        // the number of parts is calculated as bytes_this_thread / part_size rounded up.  Due to
+        // this rounding, the safe_max_parts = max_number_of_parts - number_of_client_transfer_threads + 1.
+        //
+        // Also note that because the file split per thread is deterministic, we could calculate the
+        // exact number of parts needed but this would replicate the math done in
+        // determine_start_and_end_part_from_offset_and_bytes_this_thread but to keep in simple
+        // and avoid code replication, we are just taking the easy approach by using safe_max_parts.
+        // It just means the cutoff to drain the circular buffer is a small percentage
+        // below that actual necessary value so that the code is kept clean.
```
korydraughn, line 852: I take it that using `safe_max_parts` doesn't introduce any failure cases?

Is the implementation stronger by duplicating the math? To avoid duplication, couldn't we add a function for that and call it in the places where it's needed?

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -804,35 +807,105 @@ namespace irods::experimental::io::s3_transport
...
+        //   - SUCCESS() if the calculated part_size is less than or equal to max_part_size
+        //     (defined as 5 GiB in S3 standards).
+        //
+        //   - an error if the calculated part_size is greater than max_part_size.
+        //
+        //     This can happen due to the following scenarios:
+        //     - file_size > max_part_size * safe_max_parts - With the default values
+        //       this is 50 TiB which is the largest object possible in S3.
+        //     - circular_buffer_size > max_part_size.  This is a misconfiguration and is unlikely
+        //       as it means each buffer is greater than 5 GiB.
```
korydraughn, line 865: Is this worth detecting and reporting to the admin?

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -804,35 +807,105 @@ namespace irods::experimental::io::s3_transport
...
+            if (parts_at_current_size > safe_max_parts) {
+                part_size = (object_size + safe_max_parts - 1) / safe_max_parts;
+            }
+
+            if (part_size > max_part_size) {
+                return ERROR(S3_PUT_ERROR, fmt::format(
+                        "Object of size {} bytes cannot be uploaded via S3 multipart upload:  "
+                        "a part size of {} bytes would be required to stay within the {} part "
+                        "limit, which exceeds the configured maximum part size (S3_MAX_UPLOAD_SIZE_MB) "
+                        "of {} bytes.",
```
korydraughn:
```suggestion
                        "Object of size [{}] bytes cannot be uploaded via S3 multipart upload:  "
                        "a part size of [{}] bytes would be required to stay within the [{}] part "
                        "limit, which exceeds the configured maximum part size (S3_MAX_UPLOAD_SIZE_MB) "
                        "of [{}] bytes.",
```

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -1094,9 +1167,24 @@ namespace irods::experimental::io::s3_transport
...
+            // below on the very first iteration.
+            int64_t preferred_part_size = std::max(
+                    static_cast<std::int64_t>(1LL*1024*1024*1024),
+                    static_cast<std::int64_t>(
+                        (cache_file_size + constants::MAXIMUM_NUMBER_ETAGS_PER_UPLOAD - 1) / constants::MAXIMUM_NUMBER_ETAGS_PER_UPLOAD));
+
+            if (preferred_part_size > config_.max_single_part_upload_size) {
+                logger::error("{}:{} ({}) [[{}]] Cache file of size {} cannot be uploaded via S3 multipart "
+                        "upload:  the required part size of {} exceeds the configured maximum part size "
+                        "(S3_MAX_UPLOAD_SIZE_MB) of {}.",
```
korydraughn:
```suggestion
                logger::error("{}:{} ({}) [[{}]] Cache file of size [{}] cannot be uploaded via S3 multipart "
                        "upload:  the required part size of [{}] exceeds the configured maximum part size "
                        "(S3_MAX_UPLOAD_SIZE_MB) of [{}].",
```

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -1094,9 +1167,24 @@ namespace irods::experimental::io::s3_transport
...
 
-            // Preferred part size is the largest part size that will be attempted. At 1 GiB, that still allows the largest possible
-            // file size (1 TiB) to be uploaded within the 10,000 part limit imposed by AWS.
-            int64_t preferred_part_size = 1LL*1024*1024*1024;
+            // Preferred part size is the largest part size that will be attempted, starting at 1 GiB.
+            // For cache files so large that 1 GiB parts would exceed the 10,000 part limit imposed by
+            // AWS, start instead from the smallest part size that keeps the whole file within that
+            // limit (issue #2185), so we don't immediately hit the number_of_parts > 10000 bail-out
+            // below on the very first iteration.
+            int64_t preferred_part_size = std::max(
```
korydraughn, line 1175: Can be `const` if not changed later.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -1292,6 +1380,42 @@ namespace irods::experimental::io::s3_transport
 
         }  // end populate_open_mode_flags
 
+        // Computes part_size_ and draining_enabled_ for the streaming (non-cache) multipart
+        // upload path.  Must be called only after populate_open_mode_flags() (use_cache_ must
+        // be known) and only when !use_cache_ && config_.multipart_enabled.  Returns false (and
```
korydraughn:
```suggestion
        // be known) and only when (!use_cache_ && config_.multipart_enabled) is true.  Returns false (and
```

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -1292,6 +1380,42 @@ namespace irods::experimental::io::s3_transport
...
+        // sets an error) if the object cannot be uploaded under any part size.
+        bool validate_and_compute_part_size_and_set_buffer_draining_flag()
+        {
+            std::int64_t computed_part_size;
+
+            irods::error ret = compute_part_size_for_object(
```
korydraughn: const?

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -1292,6 +1380,42 @@ namespace irods::experimental::io::s3_transport
...
+                return false;
+            }
+
+            part_size_ = computed_part_size;
+            draining_enabled_ = part_size_ > static_cast<std::int64_t>(config_.circular_buffer_size);
+
+            if (draining_enabled_) {
+                logger::warn("{}:{} ({}) [[{}]] Object size {} would require more than {} parts at "
+                        "the configured circular buffer size of {} bytes.  Increasing part size to "
+                        "{} bytes and streaming with buffer draining enabled.  Failed parts will not "
```
korydraughn:
```suggestion
                logger::warn("{}:{} ({}) [[{}]] Object size [{}] would require more than [{}] parts at "
                        "the configured circular buffer size of [{}] bytes.  Increasing part size to "
                        "[{}] bytes and streaming with buffer draining enabled.  Failed parts will not "
```

README.md
```diff
@@ -132,7 +132,7 @@ An additional flag called `HOST_MODE` is used to enable cacheless mode.  The def
...
... store data while it is being streamed to S3.  The size of the circular buffer is CIRCULAR_BUFFER_SIZE * S3_MPU_CHUNK.  The default value is 4 so if the S3_MPU_CHUNK is the default of 5 MiB the circular buffer size will be 20 MiB.  CIRCULAR_BUFFER_SIZE must be at least 2.  If a size is set lower than 2 then it will default to 2.  AWS S3 multipart uploads are limited to 10,000 parts, so for a sufficiently large object the plugin will automatically grow the part size beyond the configured circular buffer size to stay within that limit.  When this happens, the circular buffer is drained as data must be erased immediately upon streaming to S3, so a failed part cannot be retried (the whole upload must be retried).  An object is rejected outright at open time if it is too large to be represented by S3 multipart upload at all, i.e. it would require a part size larger than the configured `S3_MAX_UPLOAD_SIZE_MB` even when using all 10,000 parts.  `S3_MAX_UPLOAD_SIZE_MB` defaults to 5 GiB (AWS's real per-part maximum), giving a rough ceiling of 48 TiB or more, depending on the number of parallel transfer threads; it may be raised for non-AWS S3-compatible endpoints that support larger parts.
```
korydraughn, line 135: > ... it may be raised for non-AWS S3-compatible endpoints ...

What does this mean?

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,175 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+// conditions can be exercised without needing real huge files.  Whether draining is
+// required is inferred by the caller (part_size > circular_buffer_size), so it is not
+// returned directly here.
+TEST_CASE("test_compute_part_size_for_object", "[part_splits][compute_part_size]")
+{
+    using s3_transport = irods::experimental::io::s3_transport::s3_transport<char>;
+
+    SECTION("object fits within the part limit at the configured buffer size - unchanged")
+    {
+        std::int64_t part_size = -1;
```
korydraughn: Move this line above the `SECTION` and delete it from the other `SECTION`s. That dedups some of the code and aligns with the use-case for `SECTION`s.

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,175 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+                1024*1024*1024,      // max_part_size
+                part_size);
+
+        REQUIRE(ret.ok());
+        REQUIRE(part_size == 10*1024*1024);  // unchanged - no draining required
+    }
+
+    SECTION("object requires more parts than the limit allows - part size grows")
+    {
+        std::int64_t part_size = -1;
```
korydraughn: Remove.

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,175 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+        // total parts, even with per-thread rounding, must stay within max_number_of_parts
+        std::int64_t worst_case_parts = number_of_client_transfer_threads +
+                (object_size + part_size - 1) / part_size;
+        REQUIRE(worst_case_parts <= max_number_of_parts + number_of_client_transfer_threads);
+        REQUIRE((object_size + part_size - 1) / part_size <= max_number_of_parts);
+    }
+
+    SECTION("object is too large for S3 multipart upload under any part size - fails")
+    {
+        std::int64_t part_size = -1;
```
korydraughn: Remove.

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,175 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+                1000,                // max_part_size
+                part_size);
+
+        REQUIRE_FALSE(ret.ok());
+        REQUIRE(part_size > 1000);
+    }
+
+    SECTION("unknown object size is left unchanged")
+    {
+        std::int64_t part_size = -1;
```
korydraughn: Remove.

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,175 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+TEST_CASE("test_circular_buffer_pop_front_n_array", "[circular_buffer]")
+{
+    irods::experimental::circular_buffer<char> cb{20};
+
+    std::string data = "0123456789";
+    cb.push_back(data.begin(), data.end());
+
+    char out[4] = {};
+    cb.pop_front(4, out);
+    REQUIRE(std::string(out, 4) == "0123");
```
korydraughn: Use `std::array<char, 4>` and `.size()` to avoid hard-coding the 4. Or use a `constexpr` variable.

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,175 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+
+    char out2[6] = {};
+    cb.pop_front(6, out2);
+    REQUIRE(std::string(out2, 6) == "456789");
```
korydraughn: Use `std::array` or constexp variable for 6.

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,175 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+
+    // buffer should now be empty - pushing capacity() worth of new data should all fit
+    std::string more_data(20, 'x');
+    auto inserted = cb.push_back(more_data.begin(), more_data.end());
+    REQUIRE(inserted == 20);
```
korydraughn:
```suggestion
    REQUIRE(inserted == more_data.size());
```

unit_tests/src/test_s3_transport.cpp
```diff
@@ -1199,3 +1199,175 @@ TEST_CASE("test_part_splits", "[part_splits]")
...
+        int n = cb_writer.callback_implementation(sizeof(chunk), chunk);
+        REQUIRE(n >= 0);
+        result.append(chunk, static_cast<std::size_t>(n));
+    }
+
+    producer.join();
+
+    REQUIRE(result == source);
+
+    // post_success_cleanup() must be a no-op when draining_enabled -- bytes were already
```
korydraughn:
```suggestion
    // post_success_cleanup() must be a no-op when draining_enabled is true -- bytes were already
```

korydraughn (commented):
Finished reviewing remaining files.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: I actually thought of that but didn't implement it.  Will look into it.

The `safe_max_parts` should not introduce failure cases.  It just means the cutoff is a bit lower than what is actually required.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: This is the same pattern that is done in all of the S3 tests including the cache classes that we inherited.

The idea is that things in `__init__` run once rather than on every test.  But with it doing so little that distinction is somewhat trivial.

This could be in `setUp()` but for consistency with the existing classes I'd prefer to leave it in `__init__`.

packaging/resource_suite_s3_nocache.py

packaging/resource_suite_s3_nocache.py

packaging/resource_suite_s3_nocache.py
JustinKyleJames: It is not used.  The tests only use self.user0.

packaging/resource_suite_s3_nocache.py

packaging/resource_suite_s3_nocache.py
trel: a fine candidate for general refactor once we have coverage reports everywhere.  @MartinFlores751

packaging/resource_suite_s3_nocache.py
JustinKyleJames: The resource name is appended to this directory in the S3 plugin code so I don't think that that is necessary.  Also this would make this test inconsistent with others.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: This is also copied from others which was derived from the original tests we inherited.

I believe `static_bucket_name` was intended to be if the tests all use some static bucket and do not create a bucket for each test.  In that case, you wouldn't want to remove the bucket after the test ran.

packaging/resource_suite_s3_nocache.py
korydraughn: So, something like `/var/lib/irods/<resc_name>/` is created by the plugin to provide isolation and organization?

If so, then yes I agree no additional action is necessary.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: I changed the comment mostly removing the underlines so the concept is referenced but actual variable names are not used.

I also shortened the comment.

packaging/test_irods_resource_plugin_s3.py
JustinKyleJames: I added a disk usage check but did it with `skipIf(True or psutil.disk_usage('/').free < ...`.

I think the default should be to skip for the AWS tests even if the disk is large enough.  According to AI, the cost to run this test from a non-AWS computer is approximately $9.  So the default should be to skip it.  In either case, it won't run unless the available free space is big enough.

packaging/test_irods_resource_plugin_s3_minio.py
JustinKyleJames: I did just as for AWS except it doesn't have the skip default.  If the disk is large enough it will run.

packaging/resource_suite_s3_nocache.py

packaging/resource_suite_s3_nocache.py
JustinKyleJames: That is correct.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: Do you want me to change this even though similar methods exist in other classes?

I would agree that these should probably be in a common library rather than multiple copies but for now I think the names should be consistent.

s3_transport/include/irods/private/s3_transport/callbacks.hpp
JustinKyleJames: done

s3_transport/include/irods/private/s3_transport/callbacks.hpp
JustinKyleJames: done

packaging/resource_suite_s3_nocache.py
korydraughn: Capturing this in an issue and addressing it later sounds good to me.

packaging/resource_suite_s3_nocache.py
korydraughn: Are you going to remove the bobby user?

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: I went ahead and put in a detection if this is greater than 5 GiB to set it to 5 GiB and provided a warning.  I did this outside of the transport code itself.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: done

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: done

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: done

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: done

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: done

README.md
JustinKyleJames: The maximum upload size (either non-multipart or a single part within a multipart upload) is 5 MiB.

This threshold can be configured with `S3_MAX_UPLOAD_SIZE_MB` in the resource plugin if an S3 vendor allows a larger upload or limits it to something less than 5 MiB.

This would be non-S3 compatible and I don't know if all vendors enforce this.  I believe this configuration has existed for a long time.

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: Actually this gets halved in a loop so it can't be constant.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: I removed the bobby user from the code.

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

unit_tests/src/test_s3_transport.cpp
JustinKyleJames: done

packaging/resource_suite_s3_nocache.py
JustinKyleJames: I opened an issue for this.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: I've implemented this and saved it in https://github.com/JustinKyleJames/irods_resource_plugin_s3/tree/handle_large_files_beyond_200GiB.  All tests pass.

JustinKyleJames:
After talking about the limitations (specifically the inability to retry parts) we have decided to put this off.  This work has been saved in [Handle Large Files Branch](https://github.com/JustinKyleJames/irods_resource_plugin_s3/tree/handle_large_files_beyond_200GiB).

That passes all tests including the very large file test that exercises the new code paths.

The suggestion to handle large files is to do the following:

1. Increase the circular buffer size to object_size / 10,000.
2. Do this for one S3 resource only that only handles very large objects.
3. Possibly connect this resource to a server that uses fewer threads so fewer large buffers are used per upload.
4. Route S3 uploads for objects > 200 GiB to this S3 resource.  Route all others to the normal S3 resource.  This should be doable in a rule.

For future reference, here is the basic math behind this approach:

**Precondition**

IRODS splits files into file_size / # threads (integer division) with the remaining bytes going to the last thread.

**Algorithm**

* The S3 plugin sets the part size to the circular buffer size.  It then calculates how many parts will be used.
* If this is <= 10,000, then we are good.
* If not then retries are disabled.  Part size is determined by the following: 
  * Split the 10,000 parts among all of the threads.  (P parts/thread)
  * Size each part (S) by the size of the last thread (N):  S = ceil(N/P)
  * Check if part size (S) is within part size limit:
    * If it is then this is the part distribution.
    * If not:
      * Use the maximum part size as the part size:
      * P = ceil(bytes for this thread / maximum part size).
        * Last thread may have a different P than others
      * Check the number of total parts needed.  If it is > 10,000, error out.

Note that if the number of threads exactly divides 10,000, then we can reach the 50 TiB object size using this method.

If the number of threads does not exactly divide 10,000, then this will error out before we reach the 50 TiB object size.  For example, if there are 3 threads, each thread has 3333 parts of size 5 GiB, you can add two more bytes (which will push thread 3 to 3334 parts) without an error but once you add another byte, that byte gets spread out on all three threads which will push them to 3334 parts each.  That will cause us to error out.  This is merely a consequence of not be able to share bytes between threads not a consequence of this particular implementation.

korydraughn:
Closing this PR now that the story is captured and the issue is linked.

PR #2317, opened 2026-08-21
[2234] Detect objects too large to upload early with better err msgs. (main)

JustinKyleJames:
1. Add a section in the README for best practices around handling very large files.
2. Handle large files early in resolve_resource_hierarchy prior to the object being locked.
3. Since thread count is not known during resolve_resource_hierarchy, detect files that are only too big because of byte splitting per thread in open(). This is after the object has been locked and for the multithreaded case the explicit error is not passed back to the client.
4. When MPU uploads are disabled, use the maximum single part upload file limit in these checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,9 +97,9 @@ S3_DEFAULT_HOSTNAME=s3.<bucket-region>.amazonaws.com
 
 To define S3 provider constraints and control multipart behavior:
 -   `S3_MPU_CHUNK` - This defines the minimum part size allowed (in MB).  The default is 5MB which is the minimum part size defined in AWS.
--   `S3_ENABLE_MPU=0` disables multipart uploads.
--   `S3_MAX_UPLOAD_SIZE_MB` - This defines the maximum upload size for non-multipart uploads (in MB), maximum part size, as well as the maximum size when using the CopyObject API.  The default is 5120MB (5GB).  This setting is ignored if MPU uploads are disabled.
--   `S3_MPU_THREADS` is the number of parts to upload in parallel.
+-   `S3_ENABLE_MPU=0` - Disables multipart uploads.
+-   `S3_MAX_UPLOAD_SIZE_MB` - This defines the maximum upload size for non-multipart uploads (in MB), maximum part size, as well as the maximum size when using the CopyObject API.  The default is 5120 MiB (5 GiB).
+-   `S3_MPU_THREADS` - The number of parts to upload in parallel.
 -   `S3_URI_REQUEST_STYLE` - The path request style used.  This is either "path" or "virtualhost".  The default is "path".  See [path vs virtual hosted requests](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html).
 -   `S3_STORAGE_CLASS` - Defines the Glacier storage class used to archive objects. Valid values are "Standard", "Glacier", "Deep_Archive", and "Glacier_IR". The default is "Standard", and these values are not case sensitive. See [Storage Class Intro](https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-class-intro.html).
 -   `S3_RESTORATION_DAYS` - The number of days an object is to be restored when restoring from Glacier.  See [RestoreObject API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html).
@@ -148,8 +148,6 @@ Cacheless mode has a few extra configuration parameters in addition to HOST_MODE
 -   `CIRCULAR_BUFFER_TIMEOUT_SECONDS` - The number of seconds the plugin will wait when waiting to read or write data from the circular buffer.  The default is 180s.
 -   `S3_CACHE_DIR` - This is the directory where temporary cache files are located in cases where a cache file is required.  (See below.)  The default is `/tmp`.
 
-Note that with the default values for CIRCULAR_BUFFER_SIZE and S3_MPU_CHUNK, due to the 10,000 part limit in AWS, the largest file that can be uploaded is 20 MiB * 10,000 or roughly 200 GiB.  To upload larger files these defaults need to be updated which will have memory use implications.
-
 The following is an example of how to configure a `cacheless_attached` S3 resource:
 
 ```
@@ -160,6 +158,19 @@ Some configuration settings have special meaning when the resource is in cachele
 -   When S3_ENABLE_MPU = 0, a cache file will be used when the S3 plugin receives parallel uploads from iRODS.
 -   When iRODS is using parallel transfer but each transfer part is less than S3_MPU_CHUNK, a cache file will be used
 -   The S3_MPU_THREADS setting is only used when flushing a cache file to S3.  In streaming mode iRODS controls the number of transfer threads that are used.
+
+#### Handling Very Large Objects
+
+So that part retries can be performed, the part size is limited to the circular buffer size which is CIRCULAR_BUFFER_SIZE * S3_MPU_CHUNK (in MiB). By default this is 4 * 5 MiB = 20 MiB.
+
+The object size must be less than 10,000 times this circular buffer size. Using the defaults that is 20 MiB * 10,000 or roughly 200 GiB. In cases where 10,000 is not evenly divisible by the thread count, the maximum object size will be a little less than this due to the way the bytes are split between parts.
+
+To handle larger objects, the CIRCULAR_BUFFER_SIZE must be increased. Note that the total memory used is the (circular buffer size) * (average number of threads) * (number of simultaneous transfers).
+
+To reduce the amount of memory used:
+
+1. Set up one resource for normal/smaller objects that have a smaller CIRCULAR_BUFFER_SIZE.
+2. Set up another resource for very large objects that have a larger CIRCULAR_BUFFER_SIZE. Scale this to the size needed.
 
 #### Cache Rules When Using Cacheless Mode
 
--- a/packaging/resource_suite_s3_nocache.py
+++ b/packaging/resource_suite_s3_nocache.py
@@ -1816,6 +1816,135 @@ OUTPUT ruleExecOut
             s3plugin_lib.remove_if_exists(file1)
             s3plugin_lib.remove_if_exists(file2)
 
+    def cleanup_failed_put__issue_2234(self, filename):
+        # The part-count check in resolve_hierarchy only has the object's size to work
+        # with. The number of transfer threads is not known until after hierarchy
+        # resolution. Object uploads that fail because of the size and thread count
+        # combination are rejected after the object is locked. This method removes
+        # that lock.
+        logical_path = self.user0.session_collection + '/' + filename
+        self.admin.run_icommand(['iadmin', 'modrepl', 'logical_path', logical_path,
+                'replica_number', '0', 'DATA_REPL_STATUS', '0'])
+        self.user0.run_icommand(['irm', '-f', filename])
+
+    def test_iput_fails_cleanly_when_object_needs_too_many_parts__issue_2234(self):
+        try:
+            hostname = lib.get_hostname()
+            hostuser = getpass.getuser()
+            resource_name = f'{(inspect.currentframe().f_code.co_name)[:58]}_resc'  # resource names must be <= 63 chars
+
+            # With S3_MPU_CHUNK=5 (MB) and CIRCULAR_BUFFER_SIZE=2, the fixed part size used
+            # for streaming multipart uploads is 10 MiB.  AWS S3 multipart uploads are capped
+            # at 10,000 parts, so any object that would need more than 10,000 parts will be rejected.
+            #
+            # When MPU is disabled, any object over the maximum single put size will be rejected.
+            s3_context = self.s3_context + ';S3_MPU_CHUNK=5;CIRCULAR_BUFFER_SIZE=2'
+
+            self.admin.assert_icommand(f'iadmin mkresc {resource_name} s3 {hostname}:/{self.s3bucketname}/{hostuser}/{resource_name} {s3_context}', 'STDOUT_SINGLELINE', 'Creating')
+
+            circular_buffer_size = 5 * 1024 * 1024 * 2  # S3_MPU_CHUNK(5MB) * CIRCULAR_BUFFER_SIZE(2)
+            maximum_number_of_parts = 10000
+
+            # Default S3_MAX_UPLOAD_SIZE_MB is not overridden by s3_context above.
+            max_single_part_upload_size = 5 * 1024 * 1024 * 1024  # 5 GiB
+
+            file1 = f'{inspect.currentframe().f_code.co_name}_f1'
+
+            if self.s3EnableMPU:
+                # One byte larger than the circular buffer size * maximum number of parts.
+                # This needs more than 10,000 parts no matter how the bytes are divided
+                # among threads.
+                file_size = circular_buffer_size * maximum_number_of_parts + 1
+                expected_error = 'S3_PUT_ERROR'
+                expected_message = 'cannot be uploaded via S3 multipart upload'
+            else:
+                # MPU disabled means the object is always uploaded as a single part (one S3
+                # PUT) capped at the maximum single part upload size.
+                file_size = max_single_part_upload_size + 1
+                expected_error = 'UNIX_FILE_OPEN_ERR'
+                expected_message = 'cannot be uploaded via an S3 PutObject'
+
+            s3plugin_lib.truncate_file(file1, file_size)
+
+            # Attempt a file over the maximum allowed size. These have different error codes to match
+            # existing behavior. The descriptive message is pushed onto the RE error stack and
+            # printed to stdout, while the status code shows up on stderr.
+            out, err, rc = self.user0.run_icommand(['iput', '-N', '1', '-f', '-R', resource_name, file1])
+            self.assertNotEqual(0, rc)
+            self.assertIn(expected_error, err)
+            self.assertIn(expected_message, out)
+            # rejected during hierarchy resolution, before any catalog entry was created
+            self.user0.assert_icommand(f'ils -L {file1}', 'STDERR_SINGLELINE', 'does not exist')
+
+        finally:
+            # cleanup
+            self.cleanup_failed_put__issue_2234(file1)
+            self.admin.assert_icommand(f'iadmin rmresc {resource_name}', 'EMPTY')
+            s3plugin_lib.remove_if_exists(file1)
+
+    def test_iput_fails_cleanly_when_thread_split_pushes_object_past_max_parts__issue_2234(self):
+        try:
+            hostname = lib.get_hostname()
+            hostuser = getpass.getuser()
+            resource_name = f'{(inspect.currentframe().f_code.co_name)[:58]}_resc'  # resource names must be <= 63 chars
+
+            s3_context = self.s3_context + ';S3_MPU_CHUNK=5;CIRCULAR_BUFFER_SIZE=2'
+
+            self.admin.assert_icommand(f'iadmin mkresc {resource_name} s3 {hostname}:/{self.s3bucketname}/{hostuser}/{resource_name} {s3_context}', 'STDOUT_SINGLELINE', 'Creating')
+
+            circular_buffer_size = 5 * 1024 * 1024 * 2  # S3_MPU_CHUNK(5MB) * CIRCULAR_BUFFER_SIZE(2)
+            maximum_number_of_parts = 10000
+            number_of_threads = 3
+
+            # With 3 threads, iRODS splits the object so the first two threads each get
+            # floor(object size / 3) bytes and the third (last) thread gets that plus the
+            # remainder.  Because the maximum number of parts (10,000) is not evenly divisible
+            # by 3, the largest object that still fits within the 10,000 part limit at this
+            # part size is 9999*(circular buffer size) + 2 bytes.
+            true_maximum_object_size_for_three_threads = (maximum_number_of_parts - 1) * circular_buffer_size + 2
+            file_size = true_maximum_object_size_for_three_threads + 1
+
+            file1 = f'{inspect.currentframe().f_code.co_name}_f1'
+
+            s3plugin_lib.truncate_file(file1, file_size)
+
+            # When MPU is enabled, this object's size alone does not exceed the part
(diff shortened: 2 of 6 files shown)


s3_resource/src/s3_operations.cpp
```diff
@@ -2365,6 +2392,59 @@ namespace irods_s3 {
...
+
+            std::int64_t max_object_size_for_part_count =
+                    irods::experimental::io::s3_transport::constants::MAXIMUM_NUMBER_ETAGS_PER_UPLOAD *
+                    circular_buffer_size;
+
+            if (static_cast<std::int64_t>(data_size) > max_object_size_for_part_count) {
+                const std::string msg = fmt::format(
+                        "Object of size [{}] bytes cannot be uploaded via S3 multipart upload. It "
+                        "would require more than [{}] parts at the configured circular buffer size "
+                        "(S3_MPU_CHUNK/CIRCULAR_BUFFER_SIZE) of [{}] bytes. Increase the circular "
```
trel: is this correct?

`configured circular buffer size (S3_MPU_CHUNK/CIRCULAR_BUFFER_SIZE)`

Feels like the math(?) should be different...

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
```diff
@@ -1365,6 +1389,24 @@ namespace irods::experimental::io::s3_transport
...
+            // the open needs to be rejected.
+            if (!use_cache_ && config_.multipart_enabled &&
+                    total_number_of_parts_for_object(config_.object_size,
+                            config_.number_of_client_transfer_threads,
+                            static_cast<std::int64_t>(config_.circular_buffer_size))
+                        > constants::MAXIMUM_NUMBER_ETAGS_PER_UPLOAD) {
+                this->set_error(ERROR(S3_PUT_ERROR, fmt::format(
+                        "Object of size [{}] bytes cannot be uploaded via S3 multipart upload. It would "
+                        "require more than [{}] parts at the configured circular buffer size "
+                        "(S3_MPU_CHUNK/CIRCULAR_BUFFER_SIZE) of [{}] bytes. Increase the circular buffer "
```
trel: same confusion here for me.

README.md
```diff
@@ -161,6 +159,19 @@ Some configuration settings have special meaning when the resource is in cachele
 -   When iRODS is using parallel transfer but each transfer part is less than S3_MPU_CHUNK, a cache file will be used
 -   The S3_MPU_THREADS setting is only used when flushing a cache file to S3.  In streaming mode iRODS controls the number of transfer threads that are used.
 
+#### Handling Very Large Objects
+
+So that part retries can be performed, the part size is limited to the circular buffer size which is CIRCULAR_BUFFER_SIZE * S3_MPU_CHUNK (in MiB). By default this is 4 * 5 MiB = 20 MiB.
```
trel, line 164: maybe we need another word for the total circular buffer size... rather than just the size of each 'section' or 'slot' of the buffer.

s3_resource/src/s3_operations.cpp
JustinKyleJames: Well it wasn't intended to be a formula but agree it is confusing.

README.md
JustinKyleJames: We may need to discuss this.  I agree it is confusing and is that way for historical reasons.  The original CIRCULAR_BUFFER_SIZE was in MiB but when we changed it to allow retries and forced there to be at least 2 S3_MPU_CHUNKs, I changed it to multiples of S3_MPU_CHUNK.

I would suggest deprecating CIRCULAR_BUFFER_SIZE for something like CIRCULAR_BUFFER_SIZE_IN_MPU_CHUNKS or something like that.

s3_transport/include/irods/private/s3_transport/s3_transport.hpp
JustinKyleJames: done

s3_resource/src/s3_operations.cpp
JustinKyleJames: I made the change to S3_MPU_CHUNK * CIRCULAR_BUFFER_SIZE.

JustinKyleJames:
I force pushed a small change to the original commit message.

README.md
trel: so we have...
- buffer chunk count (misleadingly, today `CIRCULAR_BUFFER_SIZE`)
- mpu chunk size (today `S3_MPU_CHUNK`)
- total buffer size (today math of `CIRCULAR_BUFFER_SIZE * S3_MPU_CHUNK`)

so new names could be...
- `CIRCULAR_BUFFER_CHUNKS`
- `S3_MPU_CHUNK_SIZE_IN_MB`
- math of `CIRCULAR_BUFFER_CHUNKS * S3_MPU_CHUNK_SIZE_IN_MB`

or something else...  maybe we need a new issue?

trel (commented):
yes, agreed, current code is good to squash.

packaging/s3plugin_lib.py
```diff
@@ -5,6 +5,10 @@ def remove_if_exists(localfile):
     if os.path.exists(localfile):
         os.unlink(localfile)
 
+def truncate_file(f_name, f_size):
+    with open(f_name, 'wb'):
+        os.truncate(f_name, f_size)
```
korydraughn: Shouldn't this be the following?

```suggestion
    with open(f_name, 'wb') as f:
        f.truncate(f_size)
```

packaging/resource_suite_s3_nocache.py
```diff
@@ -1816,6 +1816,135 @@ def test_put_get_various_file_sizes(self):
...
+            self.assertIn(expected_error, err)
+            self.assertIn(expected_message, out)
+            # rejected during hierarchy resolution, before any catalog entry was created
+            self.user0.assert_icommand(f'ils -L {file1}', 'STDERR_SINGLELINE', 'does not exist')
+
+        finally:
+
+            # cleanup
+            self.cleanup_failed_put__issue_2234(file1)
+            self.admin.assert_icommand("iadmin rmresc s3resc1", 'EMPTY')
```
korydraughn: `assert_icommand` -> `run_icommand`

Is the name `s3resc1` used by other tests? Is the name special in any way?

packaging/resource_suite_s3_nocache.py
```diff
@@ -1816,6 +1816,135 @@ def test_put_get_various_file_sizes(self):
...
+            file1 = f'{inspect.currentframe().f_code.co_name}_f1'
+
+            s3plugin_lib.truncate_file(file1, file_size)
+
+            # When MPU is enabled, this object's size alone does not exceed the part
+            # limit, only the 3-way thread split does. The number of transfer threads
+            # isn't known until after hierarchy resolution (iRODS core computes it via
+            # getNumThreads() but only after the catalog replica is already registered
+            # and locked. The client independently derives a generic SYS_COPY_LEN_ERR
+            # (from a byte-count mismatch) rather asking the server for the underlying
```
korydraughn: Should the word `than` come after `rather`?


```suggestion
            # (from a byte-count mismatch) rather than asking the server for the underlying
```

packaging/resource_suite_s3_nocache.py
```diff
@@ -1816,6 +1816,135 @@ def test_put_get_various_file_sizes(self):
...
+            out, err, rc = self.user0.run_icommand(['iput', '-N', '1', '-f', '-R', 's3resc1', file1])
+            self.assertNotEqual(0, rc)
+            self.assertIn(expected_error, err)
+            self.assertIn(expected_message, out)
+            # rejected during hierarchy resolution, before any catalog entry was created
+            self.user0.assert_icommand(f'ils -L {file1}', 'STDERR_SINGLELINE', 'does not exist')
+
+        finally:
+
+            # cleanup
```
korydraughn, line 1880:
```suggestion
        finally:
            # cleanup
```

packaging/resource_suite_s3_nocache.py
```diff
@@ -1816,6 +1816,135 @@ def test_put_get_various_file_sizes(self):
...
+            self.assertIn(expected_error, err)
+            if expected_message is not None:
+                self.assertIn(expected_message, out)
+
+            if not self.s3EnableMPU:
+                self.user0.assert_icommand(f'ils -L {file1}', 'STDERR_SINGLELINE', 'does not exist')
+
+        finally:
+
+            # cleanup
```
korydraughn, line 1943:
```suggestion
        finally:
            # cleanup
```

s3_resource/src/s3_operations.cpp
```diff
@@ -2365,6 +2392,59 @@ namespace irods_s3 {
...
+                addRErrorMsg(&_ctx.comm()->rError, S3_PUT_ERROR, msg.c_str());
+                return ERROR(S3_PUT_ERROR, msg);
+            }
+        }
+
+        // When MPU is disabled, an object is always uploaded as a single S3 PUT which has
+        // a maximum size of S3_MAX_UPLOAD_SIZE_MB (usually 5 GiB).
+        if ((irods::CREATE_OPERATION == *_opr || irods::WRITE_OPERATION == *_opr) &&
+                data_size > 0 &&
+                data_size != static_cast<std::uint64_t>(s3_transport_config::UNKNOWN_OBJECT_SIZE) &&
```
korydraughn: Looks like this set of conditions can be assigned to a variable and reused here and on L2398-L2400.

Or, you branch on `s3GetEnableMultiPartUpload(_ctx.prop_map())` since that seems to be the gating condition.

s3_resource/src/s3_operations.cpp
```diff
@@ -2365,6 +2392,59 @@ namespace irods_s3 {
...
+            std::int64_t max_object_size_for_part_count =
+                    irods::experimental::io::s3_transport::constants::MAXIMUM_NUMBER_ETAGS_PER_UPLOAD *
+                    circular_buffer_size;
+
+            if (static_cast<std::int64_t>(data_size) > max_object_size_for_part_count) {
+                const std::string msg = fmt::format(
+                        "Object of size [{}] bytes cannot be uploaded via S3 multipart upload. It "
+                        "would require more than [{}] parts at the configured circular buffer size "
+                        "(S3_MPU_CHUNK*CIRCULAR_BUFFER_SIZE) of [{}] bytes. Increase the circular "
+                        "buffer size to reduce the number of parts required.",
```
korydraughn: These messages are written to the rError stack which means users who have no idea that iRODS is in use can possibly receive this.

These messages feel more for the admin. We should definitely log this and consider  providing a more general message for the client side of things. Only the admin will understand the S3_MPU_CHUNK/CIRCULAR_BUFFER_SIZE reference.

Thoughts?

_This applies to the other C++ locations that produce similar messages._

README.md
JustinKyleJames: I can open an issue on that.

JustinKyleJames: I added issue https://github.com/irods/irods_resource_plugin_s3/issues/2318.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: It is but only created and removed within those tests themselves so there should be no conflict.

s3_resource/src/s3_operations.cpp
JustinKyleJames: I get that users might not understand the message but I want some detail somewhere.

How about I send the following back to the client?

```
Object of size [{}] bytes cannot be uploaded via S3 multipart upload. It would require more than [{}] parts at the configured buffer size.  Refer to the iRODS log for more information on this error.
```

packaging/resource_suite_s3_nocache.py

packaging/resource_suite_s3_nocache.py
JustinKyleJames: done

packaging/resource_suite_s3_nocache.py
JustinKyleJames: done

s3_resource/src/s3_operations.cpp
JustinKyleJames: I moved the `s3GetEnabledMultipartUpload()` check (with a simple else clause) to be nested inside a single if statement with the following condition:

```
if ((irods::CREATE_OPERATION == *_opr || irods::WRITE_OPERATION == *_opr) &&
                data_size > 0 &&
                data_size != static_cast<std::uint64_t>(s3_transport_config::UNKNOWN_OBJECT_SIZE))
```

packaging/resource_suite_s3_nocache.py
korydraughn: Is it possible to use a unique resource name per test?

Sharing resource names across tests means it's likely that we get cascading failures for tests sharing that resource name. Especially for situations where the test cannot clean up a resource.

s3_resource/src/s3_operations.cpp
JustinKyleJames: I changed the errors to the following and kept the original error as a `logger::warn()` message in the logs.

```
const std::string msg_to_client = fmt::format(                      
        "Object of size [{}] bytes cannot be uploaded via S3 multipart upload. It "                                                               
        "would require more than [{}] parts at the configured buffer size.  See "                                                             
        "the warning in the iRODS logs for more detail on this error.",                                                             
        data_size,                          
        irods::experimental::io::s3_transport::constants::MAXIMUM_NUMBER_ETAGS_PER_UPLOAD);
```

```
const std::string msg_to_client = fmt::format(                      
        "Object of size [{}] bytes cannot be uploaded because S3 multipart upload is "                                                        
        "disabled and the object is larger than the configured maximum single-part "                                  
        "upload size.  See the warning in the iRODS logs for more details on this error.",                                                                         
        data_size);
```

packaging/s3plugin_lib.py
JustinKyleJames: done

s3_resource/src/s3_operations.cpp
korydraughn: > I get that users might not understand the message but I want some detail somewhere.

Wouldn't the log message cover that?

Side note - the log message needs to include the logical path and resource. It might be helpful to include the physical path too.

> How about I send the following back to the client?
> 
> ```
> Object of size [{}] bytes cannot be uploaded via S3 multipart upload. It would require more than [{}] parts at the configured buffer size.  Refer to the iRODS log for more information on this error.
> ```

That messaging is better, but it breaks the abstraction by mentioning physical storage details.

Then again, clients already receive error codes like `S3_PUT_ERROR`. So maybe it's fine.

JustinKyleJames: I don't think there is any way around breaking the abstraction by mentioning storage details.  This is failing because of those details.  Whether the person doing the iput cares is another matter.  We can't just always pretend that we are talking to a unixfilesystem.

JustinKyleJames: What do you think about just saying the following:

`Object of size [{}] bytes cannot be uploaded via S3 multipart upload with the current configuration.  See the warning message in the iRODS log for more detail.`

It still has some detail but about as minimal as possible IMO.

Also, for the log message I will add the resource and path.

korydraughn: Yes, let's use that message.

JustinKyleJames: Here is a breakdown of the logs.

What is sent back to the client:

```
2026-08-26T19:48:27.052Z pid:1997 [rsDataObjPut_impl:651] - [S3_PUT_ERROR: [-]\t/irods_source/server/core/src/irods_resource_redirect.cpp:157:irods::error (anonymous namespace)::request_vote_for_file_object(rsComm_t *, const std::string &, const std::string &, irods::file_object_ptr, std::string &, float &) :  status [S3_PUT_ERROR]  errno [] -- messag
e [failed in call to redirect host [1f36a049d707]  hier [] vote [0]]\t[-]\t/irods_plugin_source/s3_resource/src/s3_operations.cpp:2428:irods::error irods_s3::s3_resolve_resc_hier_operation(irods::plugin_context &, const std::string *, const std::string *, irods::hierarchy_parser *, float *) :  status [S3_PUT_ERROR]  errno [] -- message [Object of size
 [104857600001] bytes cannot be uploaded via S3 multipart upload with the current configuration.  See the warning message in the iRODS log for more details.]]
```

The more detailed error in the logs:

```
2026-08-26T19:48:27.050Z pid:1997 Object [/tempZone/home/alice/2026-08-26Z19:48:21--irods-testing-8veeu7kj/test_iput_fails_cleanly_when_object_needs_too_many_parts__issue_2234_f1] of size [104857600001] bytes cannot be uploaded via S3 multipart upload. It would require more than [10000] parts at the configured circular buffer size (S3_MPU_CHUNK*CIRCULAR_BUFFER_SIZE) of [10485760] bytes. Increase the circular buffer size on resource [s3resc1] to reduce the number of parts required.
```

Note that the above log entries have been modified to appear like the older log format.

JustinKyleJames: When multipart is disabled here are the errors.

What is sent back to the client:

```
2026-08-26T19:49:06.386Z pid:2421 [-]\t/irods_source/server/core/src/irods_resource_redirect.cpp:157:irods::error (anonymous namespace)::request_vote_for_file_object(rsComm_t *, const std::string &, const std::string &, irods::file_object_ptr, std::string &, float &) :  status [UNIX_FILE_OPEN_ERR]  errno [] -- message [failed in call to redirect host 
[1f36a049d707]  hier [] vote [0]]\t[-]\t/irods_plugin_source/s3_resource/src/s3_operations.cpp:2453:irods::error irods_s3::s3_resolve_resc_hier_operation(irods::plugin_context &, const std::string *, const std::string *, irods::hierarchy_parser *, float *) :  status [UNIX_FILE_OPEN_ERR]  errno [] -- message [Object of size [5368709121] bytes cannot be
 uploaded via an S3 PutObject.  See the warning message in the iRODS log for more details.]
```

The more details error in the logs:

```
2026-08-26T19:49:06.385Z pid:2421 Object [/tempZone/home/alice/2026-08-26Z19:49:00--irods-testing-g5_8vylg/test_iput_fails_cleanly_when_object_needs_too_many_parts__issue_2234_f1] of size [5368709121] bytes cannot be uploaded because S3 multipart upload is disabled and the object is larger than the configured maximum single-part upload size (S3_MAX_UP
LOAD_SIZE_MB) of [5368709120] bytes. Enable multipart upload or increase the maximum upload size on resource [s3resc1] - if the S3 provider allows larger single part uploads - to upload this object.
```

JustinKyleJames: I also made a slight change to the README as I saw something that wasn't correct.

packaging/resource_suite_s3_nocache.py
JustinKyleJames: I used the test suite name (truncated as the resource name must be <= 63 chars) with "_resc" as the resource name.

JustinKyleJames:
This is ready and passes test.

README.md
```diff
@@ -98,7 +98,7 @@ S3_DEFAULT_HOSTNAME=s3.<bucket-region>.amazonaws.com
 To define S3 provider constraints and control multipart behavior:
 -   `S3_MPU_CHUNK` - This defines the minimum part size allowed (in MB).  The default is 5MB which is the minimum part size defined in AWS.
 -   `S3_ENABLE_MPU=0` disables multipart uploads.
```
korydraughn:
```suggestion
-   `S3_ENABLE_MPU=0` - Disables multipart uploads.
```

README.md
```diff
@@ -98,7 +98,7 @@ S3_DEFAULT_HOSTNAME=s3.<bucket-region>.amazonaws.com
...
--   `S3_MAX_UPLOAD_SIZE_MB` - This defines the maximum upload size for non-multipart uploads (in MB), maximum part size, as well as the maximum size when using the CopyObject API.  The default is 5120MB (5GB).  This setting is ignored if MPU uploads are disabled.
+-   `S3_MAX_UPLOAD_SIZE_MB` - This defines the maximum upload size for non-multipart uploads (in MB), maximum part size, as well as the maximum size when using the CopyObject API.  The default is 5120MB (5GB).
 -   `S3_MPU_THREADS` is the number of parts to upload in parallel.
```
korydraughn:
```suggestion
-   `S3_MPU_THREADS` - The number of parts to upload in parallel.
```

README.md
```diff
@@ -98,7 +98,7 @@ S3_DEFAULT_HOSTNAME=s3.<bucket-region>.amazonaws.com
 To define S3 provider constraints and control multipart behavior:
 -   `S3_MPU_CHUNK` - This defines the minimum part size allowed (in MB).  The default is 5MB which is the minimum part size defined in AWS.
 -   `S3_ENABLE_MPU=0` disables multipart uploads.
--   `S3_MAX_UPLOAD_SIZE_MB` - This defines the maximum upload size for non-multipart uploads (in MB), maximum part size, as well as the maximum size when using the CopyObject API.  The default is 5120MB (5GB).  This setting is ignored if MPU uploads are disabled.
+-   `S3_MAX_UPLOAD_SIZE_MB` - This defines the maximum upload size for non-multipart uploads (in MB), maximum part size, as well as the maximum size when using the CopyObject API.  The default is 5120MB (5GB).
```
korydraughn: Are the byte suffixes intentionally MB and GB? I remember we merged a PR recently that changed a few to use MiB and GiB.

I think you also added a space between the number and suffix too - e.g. 5 GB, 5120 MB.

See this commit - https://github.com/irods/irods_resource_plugin_s3/commit/dee867629c1a6f1540605261800922e0ac3bccf1.

korydraughn (commented):
Looks good overall.

Spotted a couple things in the README related to consistency.

README.md
trel: agreed for the docs.    but let's leave the actual setting as `_MB` though.

README.md
JustinKyleJames: done

README.md
JustinKyleJames: done

README.md
JustinKyleJames: I made the change to MiB and GiB and added a space between the value and unit.

korydraughn:
All tests passed in github actions.

Squash it.

JustinKyleJames:
Squash was done

korydraughn (approved):
Pound it.

JustinKyleJames:
> Pound it.

done

PR #2321, opened 2026-09-02
[2319] Fix shared memory leak cases (4-3-stable)

JustinKyleJames:
1. Small write into a large existing object - number_of_client_transfer_threads was computed from the existing object's size rather than the actual write, so a forced single-threaded/cache-based write seeded the counter too high and it never reached zero. Fixed by forcing number_of_client_transfer_threads = -1 when the write isn't a put_repl_flag operation.

2. Checksum read after iput -K - iRODS reuses the PUT's L1desc entry for the post-write checksum read, leaving oprType == PUT_OPR even though it's really a single-threaded read. get_number_of_threads_data_size_and_opr_type() set threads_remaining_to_close from the original PUT's thread count, but only one close() ever ran for it, leaving the counter stuck non-zero. Fixed by detecting this case via open_mode at both open time (skip reseeding) and close time (skip decrementing).

3. per_thread_data now captures oprType at open time instead of reading it from a shared global at close time, since another thread's close() could reset that global to -1 first in multi-threaded PUTs.

Assertions were added temporarily to tearDown() to validate that no tests leaked shared memory. After running that, the assertions were moved to the individual tests that originally failed and tests were re-run. All passed.


s3_resource/src/s3_operations.cpp
```diff
@@ -399,7 +405,12 @@ namespace irods_s3 {
             // If this is GET_OPR, we do not need the shared memory. Set the threads_remaining_to_close to 0 so the shmem will be
             // deleted immediately. Note that for GET_OPR we don't necessarily know the number of threads (nor do we need it) and
             // this makes it hard to determine when the shared memory can be deleted.
-            if (oprType == GET_OPR) {
+            //
+            // The same applies to a read-after-write for a checksum operation. In that case oprType is still PUT_OPR
+            // (iRODS does not update it) but open_mode shows this is actually a single read, not the original
+            // multi-threaded PUT.
+            bool is_read_after_write_for_checksum = oprType == PUT_OPR && !(open_mode & std::ios_base::out);
```
korydraughn:
```suggestion
            bool is_read_after_write_for_checksum = (oprType == PUT_OPR) && !(open_mode & std::ios_base::out);
```

s3_resource/src/s3_operations.cpp
```diff
@@ -1128,7 +1143,17 @@ namespace irods_s3 {
...
+            // Issue 2319: If oprType is -1 (unknown) do not run this code as it will recreate
+            //   shared memory and decrement threads_remaining_to_close to -1.
+            //
+            // data.oprType (captured per-fd at open, see per_thread_data above) is used here rather than
+            // the irods_s3::oprType global, since the global can already have been reset to -1 by another
+            // thread's close() by the time this one runs.
+            //
+            // The same reasoning applies to a read-after-write for a checksum operation. oprType is still
+            // PUT_OPR, but this must be treated like a GET_OPR.
+            bool is_read_after_write_for_checksum = data.oprType == PUT_OPR && !(data.open_mode & std::ios_base::out);
```
korydraughn:
```suggestion
            bool is_read_after_write_for_checksum = (data.oprType == PUT_OPR) && !(data.open_mode & std::ios_base::out);
```

s3_resource/src/s3_operations.cpp
JustinKyleJames: done

s3_resource/src/s3_operations.cpp
JustinKyleJames: done

packaging/resource_suite_s3_nocache.py
```diff
@@ -2638,6 +2642,11 @@ def test_multiple_irsync_to_s3__issue_2204(self):
             assert_command(['diff', '-q', filename1, get_filename1])
             assert_command(['diff', '-q', filename1, get_filename2])
 
+            # Make sure shared memory is cleaned up.  irsync's checksum verification opens and closes
+            # the file without setting up the L1desc table and oprType remains unknown (-1).  This
+            # caused a shared memory leak.
```
trel, line 2647:
```suggestion
            # had caused a shared memory leak.
```

packaging/resource_suite_s3_nocache.py
```diff
@@ -2695,6 +2704,11 @@ def test_multiple_irsync_to_s3_with_N_flag__issue_2204(self):
...
```
trel, line 2709:
```suggestion
            # had caused a shared memory leak.
```

packaging/s3plugin_lib.py
```diff
@@ -1,5 +1,25 @@
+import glob
 import os
 import random
+import time
+
+def get_leaked_s3_shared_memory_files():
+    return glob.glob('/dev/shm/*s3*')
```
trel, line 7: could pick up ... other things?

can this be tightened?   perhaps `*irods_s3*` or `irods_s3-shm-*`, same as the PREFIX?

however, it's just test code (designed to be run on a machine alone), so probably not required to change here.

packaging/resource_suite_s3_nocache.py

packaging/resource_suite_s3_nocache.py

packaging/s3plugin_lib.py
JustinKyleJames: Changing it to `irods_s3` as both shared memory instances begin with that string.

korydraughn:
Does the latest commit (d2ff93b) still cleanup as intended?

JustinKyleJames:
> Does the latest commit ([d2ff93b](https://github.com/irods/irods_resource_plugin_s3/commit/d2ff93bb5251195edd0501ab10615d61139cdd66)) still cleanup as intended?

Yes.  But to be clear the glob change is just in tests.

korydraughn:
Squash it.

JustinKyleJames:
> Squash it.

Done

korydraughn:
In the commit message body for item (2), should there be a word between `stuck` and `non-zero`?

```
... leaving the counter stuck non-zero.
```

JustinKyleJames:
> In the commit message body for item (2), should there be a word between `stuck` and `non-zero`?
> 
> ```
> ... leaving the counter stuck non-zero.
> ```

I changed the wording to "so the open counter never reached zero."

korydraughn (approved):
Pound it.

JustinKyleJames:
> Pound it.

I forgot to add a comment that it has been pounded.

korydraughn:
Merging.

korydraughn:
Please cherry-pick to the main branch.

